PR #2, opened 2021-01-08
Reader for biologic mpt

ScottSoren:
I've written a Reader class for biologic mpt, `ixdat.readers.biologic.BiologicMPTReader`.
It successfully reads the test .mpt file (provided by @kkrempl in July) in **ixdat/test_data/biologic_mpt_and_zilien_tsv/2020-07-29 10_30_39 Pt_poly_cv_01_02_CVA_C01.mpt**, as demonstrated by the test script **ixdat/tests/test_biologic_reader.py**
This test script also demonstrates how the reader is invoked via `Measurement`, and how the measurement's `tstamp` is just as dynamic as planned.
More work to be done on ECMeasurement, but I'm making this pull request right away to have an example when writing other Readers, and because there were a few tiny bugs in the ixdat core that we should get fixed on master before making any other branches. 
@KennethNielsen and @kkrempl, I don't anticipate much review work, but do check if there's anything that jumps out that you'd like to see done better before giving approval. Thanks!


src/ixdat/readers/biologic.py
```diff
@@ -1,9 +1,212 @@
...
+        """Split the line and append the numbers the corresponding data column arrays"""
+        data_strings_from_line = line.strip().split()
+        for name, value_string in zip(self.column_names, data_strings_from_line):
+            try:
+                value = float(value_string)
+            except ValueError:
+                if "," in value_string:  # oh my god, why?!
+                    value_string = value_string.replace(",", ".")
+                if "E" in value_string:  # Biologic uses capital E for sci. notation.
+                    value_string = value_string.replace(",", ".")
```
KennethNielsen, line 179: The replace is the same as above. On purpose?

tests/test_biologic_reader.py
```diff
@@ -0,0 +1,24 @@
+from pathlib import Path
+from matplotlib import pyplot as plt
+from ixdat.measurements import Measurement
+
+test_data_dir = Path(
+    "../test_data/biologic_mpt_and_zilien_tsv"
```
KennethNielsen, line 6: I think is a little vulnerable, since it requires you to be in the folder when you run the file. I think it is better to do the path relative to the script path

tests/test_biologic_reader.py
```diff
@@ -0,0 +1,24 @@
...
+
+t, v = ec_measurement.get_potential(tspan=[0, 100])
+
+ec_measurement.tstamp -= 20
+t_shift, v_shift = ec_measurement.get_potential(tspan=[0, 100])
+
+fig, ax = plt.subplots()
+ax.plot(t, v, "k", label="original tstamp")
+ax.plot(t_shift, v_shift, "r", label="shifted tstamp")
+ax.legend()
```
KennethNielsen, line 24: There should maybe be a new-line at the end

tests/test_biologic_reader.py
```diff
@@ -0,0 +1,24 @@
+from pathlib import Path
```
KennethNielsen, line 1: I think there is something here, about the structure of tests vs. "module demo" which is what I would refer to this as.

I would like to reserve the "tests" folder for software tests run by pytest, and the data files to be used for the software tests (which should be kept as small as possible for small test run times) to be kept somewhere under the "tests" folder.

As for this module. If you intend it to be a demo for users to see, then I would suggest a "demo" folder under ixdat, so that they can be run programmatically from the installed package (and the data should then be kept somewhere where it can be included in the package). If, however, this serves merely as a the script that can be conveniently run during development (or further development) or the Reader, then I think it might as well live in the `if __name__ == "__main__"` part of the reader module.

Please note, that I know that you had some ideas about including data and demos and such and I don't want to interfere too much with that as I'm sure that will be sensible. I mainly just want to keep the tests folder for software tests.

src/ixdat/readers/biologic.py
```diff
@@ -1,9 +1,212 @@
...
+        self.place_in_file = "header"
+        self.header_lines = []
+        self.timestamp_string = None
+        self.tstamp = None
+        self.ec_technique = None
+        self.N_header_lines = None
+        self.column_names = []
+        self.column_data = {}
+
+    def read(self, path_to_file):
```
KennethNielsen, line 65: It seems mostly like a reader, such as this, could (maybe should) be a function and the object here serves only the purpose of maintaining parser state. And in reality, the main thing needed to turn it into a function, would be that the data containers (and possibly parser state) would need to be passed along to the sub-parser functions as arguments. So I guess that the real review comment is for you to consider for a second whether a function based implementation might in fact be "cleaner" and more testable. What you loose, I guess ,is the ability for the resulting object to refer back to the parser and view e.g. the filesname and the raw header lines and such. And this may be important enough to keep it like it is.

In any case, if you do keep it I would suggest to make the parser object "single-use", by having a `used` instance variable (or something with a better name) which is set at the end of `read` and checked at the start of `read`, to prevent a user accidentally using it more that once, which is sure to produce weird results.

KennethNielsen (changes requested):
Looks great. I few, mostly structural, comments.

src/ixdat/readers/biologic.py
ScottSoren: Nope. Should have been replace "E" with "e". That it works with this error means that `float(float_as_str)` works fine if float_as_str has "E". Deleted the second if and it works fine.

tests/test_biologic_reader.py
ScottSoren: I don't understand, isn't this relative to the script path as it is?

ScottSoren: Ah, nevermind, I see. Yes. Changed to 
test_data_dir = `Path(__file__) / "../test_data/biologic_mpt_and_zilien_tsv"`

tests/test_biologic_reader.py
ScottSoren: fixed.

tests/test_biologic_reader.py
ScottSoren: Yes, I intend it to be readily available for anyone who might want to develop this reader - and that is how I used it, running it again and again while writing the reader. I guess then if `__name__ == __main__` is a fine place. Will move after doing the rest of the stuff in this review.

src/ixdat/readers/biologic.py
ScottSoren: I think the Reader class setup is really nice. In addition to your reasons, I'll add that this was the easiest and most pleasant writing of a file parser I have ever experienced.

Ive added the `used` attribute, refactored to `file_has_been_read`, to avoid reading the file twice.

ScottSoren: Code representing this change is in the start of `read()`:
```
        if self.file_has_been_read:
            print(f"This {self.__class__.__name__} has already read {self.path_to_file}."
                  " Returning the measurement resulting from the original read. "
                  "Use a new Reader if you want to read another file.")
            return self.measurement
```
The behavior of attributes `file_has_been_read` and `measurement` are both documented in the class docstring.

tests/test_biologic_reader.py
ScottSoren: ... and now to `Path(__file__).parent.parent.parent.parent / "test_data/biologic_mpt_and_zilien_tsv"`
since the code has been moved to the reader module in src/ixdat/readers.
This strikes me also as a bit fragile. Is there a good way to reference the project folder from within a project module?

tests/test_biologic_reader.py
ScottSoren: Shit, doesn't work. 
Are module demos incompatible with relative imports?
@KennethNielsen, I need your help before I can move forward on this. Trying to run the biologic module gives the following error:
```
Traceback (most recent call last):
  File "<input>", line 1, in <module>
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.2\plugins\python-ce\helpers\pydev\_pydev_bundle\pydev_umd.py", line 197, in runfile
    pydev_imports.execfile(filename, global_vars, local_vars)  # execute the script
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.2\plugins\python-ce\helpers\pydev\_pydev_imps\_pydev_execfile.py", line 18, in execfile
    exec(compile(contents+"\n", file, 'exec'), glob, loc)
  File "C:/Users/scott/git/ixdat/src/ixdat/readers/biologic.py", line 7, in <module>
    from . import TECHNIQUE_CLASSES
  File "C:\Program Files\JetBrains\PyCharm Community Edition 2020.3.2\plugins\python-ce\helpers\pydev\_pydev_bundle\pydev_import_hook.py", line 21, in do_import
    module = self._system_import(name, *args, **kwargs)
ImportError: attempted relative import with no known parent package
```

ScottSoren:
@KennethNielsen , 
Thanks for the review!
I implemented all the other comments but not the moving of the code from tests to a module demo since I don't know whether or how to make module demos compatible with realative imports. I'll wait for your input on that.

tests/test_biologic_reader.py
ScottSoren: .. and changed back again. See https://github.com/ixdat/ixdat/pull/2#discussion_r554450371

tests/test_biologic_reader.py
ScottSoren: Moved instead to **ixdat/development_scripts**. Resolving.

tests/test_biologic_reader.py
KennethNielsen: That's a good question. I don't really have a good answer, maybe you can import the toplevel ixdat module and ask for its __file__, but that is somehow actually the same, except that you have the "dots" at the import instead of at the path.

tests/test_biologic_reader.py
ScottSoren: ... and now, after our conversation on slack, moved back to the module under the 
`if __name__ == "__main__"` block. That block starts with this:
```
    """To run this module demo in PyCharm, configure it to run with 
    Module name = ixdat.readers.biologic,
    and *not* Script path = ...
    """
```

tests/test_biologic_reader.py
ScottSoren: ... and back again, now that it works in the module. ;)

ScottSoren:
Thanks for the chat on slack! That resolved the remaining issue. I think this is now ready to merge, pending approval. @KennethNielsen , @kkrempl

KennethNielsen (approved):
Looks great

KennethNielsen:
Approved from me.

kkrempl (approved):

KennethNielsen:
:cartwheeling:

PR #4, opened 2021-01-24
Deconvolution

kkrempl:
Here the implementation of the Deconvolution class (inherited from ECMSMeasurement) together with the .pkl reader that allows to import legacy EC-MS pickles into ixdat as well as further functionality required for datatreatment of deconvolution measurements. It's definetly not free of bugs yet but ready to review in shape and form. I hope to eliminate most of the bugs while I set up the repo and code for the deconvolution paper. Looking forward to you thoughts and suggestions!


ScottSoren (commented):
Seems it doesn't actually incorporate your latest changes in a way that I can review the most updated version all at once. I'm stuck reviewing from when you requested the review.

So first a comment on this: Great! Syntax is now fixed, and I can run it to import one of my EC_MS pickles, so it's functional :D Congrats.

I will now get back to the bigger review.

ScottSoren:
So first a comment on this: Great! Syntax is now fixed, and I can run it to import one of my EC_MS pickles, so it's functional :D Congrats. :D 

Seems it doesn't actually incorporate your latest changes in a way that I can review the most updated version all at once. I'm stuck reviewing from when you requested the review. I will now get back to the bigger review, which will nonetheless be a request for changes.

src/ixdat/techniques/deconvolution.py
```diff
@@ -0,0 +1,254 @@
...
+            area = np.trapz(kernel, time)
+            kernel = kernel/area
+
+        if matrix:
+            kernel = np.tile(kernel, (len(kernel), 1))
+            i = 1
+            while i < len(time):
+                kernel[i] = np.concatenate((kernel[0][i:], kernel[0][:i]))
+                i=i+1
+
```
ScottSoren, line 253: I would cache the kernel, by including a line
`self._kernel = kernel` in this method,
and then adding a property which accesses this and calculates it if needed 
```
@property 
def kernel(self):
    """The numerical deconvolution kernel (impulse response function), vs self.t"""
    if not self._kernel:
        self.calculate_kernel()   # renamed from gen_kernel
    return self._kernel
```
Also, store the time used for the kernel (`self.t = t_kernel  # renamed from 'time'`), and this simplifies the plotting method:
```
def plot_kernel(self):
    ...
    kernel = self.kernel
    ax.plot(self.t, kernel)
    ...
```

ScottSoren: A more radical (but not too difficult) makeover that we should consider is for `Kernel` to inherit from `Measurement` and store its `t` and `kernel` as series - this would give it access to the backend for saving and loading. Do you think that would be useful?

src/ixdat/techniques/ms.py
```diff
@@ -1,7 +1,57 @@
...
+            return time, value - np.average(bg)
+
+
+
+    def get_calib_signal(self, signal_name, tspan=None, t_bg=None):
+        """Returns a calibrated signal for a given signal name. Only works if
+        calibration dict is not None.
+
+        Args:
+            signal_name (str): Name of the signal.
```
ScottSoren, line 46: Ah, but shouldn't it now be the name of the molecule?

src/ixdat/techniques/ec_ms.py
```diff
@@ -6,3 +6,6 @@
 
 class ECMSMeasurement(ECMeasurement, MSMeasurement):
```
ScottSoren, line 7: Yes! Yes! Yes! Just tested and this inheritence works as intended. 
The ECMSMeasurement returned by your reader has both the methods `get_potential` from ECMeasurement and `get_calib_signal` and they both work as intended!

src/ixdat/techniques/ms.py
```diff
@@ -1,7 +1,57 @@
...
 
 class MSMeasurement(Measurement):
     """Class implementing raw MS functionality"""
+
+    def __init__(self, name, **kwargs):
+        """Initializes a MS Measurement
+
+        Args:
+            name (str): The name of the measurement
+            calibration (dict): calibration constants whereby the key
```
ScottSoren, line 15: Can you put a TODO here reminding any reader that this is not the final calibration implementation?
Even for this stop-gap measure, I would kind of like calibration to also bridge from mass names "M2" to molecule names "H2". I guess it would have to be in the form `{"H2": {"M2": 1.95}}` or similar, or the information that the M2 signal is for H2 needs to be stored somewhere else.
I'm happy to add a bit so that it's what I think is the minimum acceptable, without re-writing what I hope a certain company will let us use.

src/ixdat/techniques/deconvolution.py
```diff
@@ -0,0 +1,254 @@
...
+            EC_data = np.array([x_curr, y_curr, x_pot, y_pot])
+            )
+
+
+        return kernel
+
+
+
+class Kernel:
+    """
```
ScottSoren, line 97: docstring format. Also, I think the potential of this class is not fulfilled, see my comments under the `gen_kernel` method.

src/ixdat/techniques/deconvolution.py
```diff
@@ -0,0 +1,254 @@
...
+        else:
+            print('Generating blank kernel')
+            self.type = None
+
+        self.params = parameters
+        self.MS_data = MS_data
+        self.EC_data = EC_data #x_curr, y_curr, x_pot, y_pot
+
+    @property
+    def sig_area(self):
```
ScottSoren, line 144: This method is an argument for `Kernel` inheriting from `Measurement`, as `Measurement` should have this method. A discussion point for later.

src/ixdat/techniques/deconvolution.py
```diff
@@ -0,0 +1,254 @@
...
+            len(int): Timelength in seconds for which the kernel/impulse response is
+                calculated.
+            norm (bool): If true the kernel/impulse response is normalized to its
+                area.
+            matrix (bool): If true the circulant matrix constructed from the kernel/
+                impulse reponse is returned.
+        """
+        if self.type is 'functional':
+
+            time = np.arange(0,len,dt)
```
ScottSoren, line 218: I don't like the name `time` for a numpy array. It indicates to me something grander, like the `time` module. 
`t` is in my opinion perfect for this, with an underscore and a discription whenever that could be ambiguous. Here it could be `t_kernel`.

src/ixdat/techniques/deconvolution.py
```diff
@@ -0,0 +1,254 @@
...
+        dt=0.1,
+        len=100,
+        norm=True,
+        matrix=False
+        ):
+        """Generates a kernel/impulse response based on the defined mass
+        transport parameters contained in dict.
+
+        Args:
+            dt (int): Timestep for which the kernel/impulse response is calculated.
```
ScottSoren, line 208: Can you add hints for the user on how long they should make dt and len? Also, if they have a vector of time values ready, can they just use that?

src/ixdat/techniques/deconvolution.py
```diff
@@ -0,0 +1,254 @@
...
+
+        Args:
+            parameters (dict): Dictionary containing the mass transport
+                parameters with the following keys:
+                    D: Diffusion constant in liquid
+                    L: Working distance between electrode and gas/liq interface
+                    V: Gas sampling volume of the chip
+                    V_dot: Volumetric capillary flow
+                    kH: Dimensionless Henry volatility
+                MS_data (list): List of numpy arrays containing the MS signal
```
ScottSoren, line 119: avoid capitalizing variable names other than classes and constants (see [PEP 8](https://www.python.org/dev/peps/pep-0008/)). And why not just let the caller of `Kernel.__init__` just tell Kernel which `Measurement` is in use, and the names of the `VSeries` to use?

src/ixdat/techniques/deconvolution.py
```diff
@@ -0,0 +1,254 @@
...
+                    data.
+                EC_data (list): List of numpy arrays containing the EC (time,
+                    current, potential).
+        """
+
+        if MS_data is not None and parameters:
+            raise Exception(
+                'Kernel can only be initialized with data OR parameters, not both'
+                )
+        if EC_data is not None and MS_data is not None:
```
ScottSoren, line 129: I wonder if this kind of switching in the `__init__` means there are really two things we call `Kernel` that share some features but are different classes. 
Maybe there should be a `MeasuredKernel` and a `CalculatedKernel` that both inherit from `Kernel`. 
Do I understand right that a measured kernel is something you would use if you've done an impulse response experiment, and thus have data that should correspond to a kernel?
At the very least, the docstring needs to go into more detail of when you'd use parameters and when you'd use data.

src/ixdat/techniques/deconvolution.py
```diff
@@ -0,0 +1,254 @@
+"""Module for analysis of EC-MS measurements with impulse response and
+deconvolution of mass transport effects to obtain partial current densities."""
+
+from .ec_ms import ECMSMeasurement
+
```
ScottSoren, line 5: you've forgotten to import numpy and fft and matplotlib and whatever else you use.

src/ixdat/techniques/deconvolution.py
```diff
@@ -0,0 +1,254 @@
...
+            snr (int): signal-to-noise ratio used for Wiener deconvolution.
+            """
+
+
+        t_sig, v_sig = self.get_calib_signal(signal_name,
+            tspan=tspan,
+            t_bg=t_bg)
+
+        kernel = kernel_obj.generate_kernel(dt=t_sig[1]-t_sig[0])
+        kernel = np.hstack((kernel, np.zeros(len(sig) - len(kernel))))
```
ScottSoren, line 42: It looks here like kernel is calculated, mutated, used to make H, and neither are saved anywhere. 
I would have both the kernel values and H stored in the `kernel_obj` and accessed as properties. As mentioned in the comments to the `Kernel` class, the code would be more intuitive and powerful if you put more functionality into that class.

src/ixdat/readers/ec_ms.py
```diff
@@ -1,9 +1,73 @@
 from . import TECHNIQUE_CLASSES
+import pickle
```
ScottSoren, line 2: the module name clashes with the technique module name. They're in different places in ixdat but it could get confusing at some point. Maybe this one could be called `ec_ms_pkl`.

src/ixdat/readers/ec_ms.py
```diff
@@ -1,9 +1,73 @@
...
+            ".pkl" extension.
+        """
+        with open(file_path, "rb") as f:
+            data = pickle.load(f)
+
+        cols_str = data['data_cols']
+        cols_list = []
+
+        for col in cols_str:
+            if col[0] == "M" and col[-1] == "x":
```
ScottSoren, line 36: Drop the test for starting with "M" and use `endswith("-x")`, for reasons described below in more detail below when you deal with "-y"

ScottSoren (changes requested):
Hey @kkrempl , great work with this!
I've tried out the importing, and it works! Especially happy that the resulting ECMSMeasurement object has working methods form both ECMeasurement and MSMeasurement, as we hoped.
There are some style conflicts. I have only mentioned a few of them (docstrings) in the comments, because I think you should just run [black](https://black.readthedocs.io/en/stable/) on the code and then we can skip a bunch of not-so-interesting discussion about asthetics ;)
There's a couple easy but important comments on improving the pickle importer. 
Then there are lots of comments on the deconvolution module. I basically think it should take much more advantage of object orientation. One problem is that the computationally somewhat slow bits are not saved anywhere as far as I can tell, so if I want to plot a kernel and then use it it will calculate it twice. That should at a minimum be improved. But I also don't have a great idea for how it should be.
Take a look at the comments to that module and see which suggestions you agree with.

I'd like to get this merged quickly, so that I don't make a conflicting ECMSMeasurement on my branch. So just add TODO's in comments on any issue that will take some thinking about. 
A minimum to merging, though, I think, is that it passes the flake8 linter that Kenneth set up (on another branch) for pre-push hook. Don't use a ton of time setting it up if you haven't (it's on me to improve docs for the tools at the moment) - if you use black to format your code, there will probably be only a few trivial things left.

src/ixdat/readers/ec_ms.py
```diff
@@ -1,9 +1,73 @@
...
+                    data[col], data["tstamp"]
+                    ))
+
+            if col == "time/s":
+                cols_list.append(TimeSeries(
+                    col, "s",
+                    data[col], data["tstamp"]
+                    ))
+
+        measurement = Measurement('tseries_ms',
```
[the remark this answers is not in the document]
kkrempl: I would need to append to the cols list and to the cols dict seperately, so it saves a little bit of code when using the dummy measurement object. Is there any reason not to do it this way?

src/ixdat/readers/ec_ms.py
```diff
@@ -1,9 +1,73 @@
...
+                tseries=measurement[col[:-1] + 'x']
+                ))
+            if col == 'Ewe/V' or col == 'I/mA':
+                cols_list.append(ValueSeries(
+                col, "A",
+                data[col],
+                tseries=measurement['time/s']
+                ))
+
+        measurement = ECMSMeasurement(file_path,
```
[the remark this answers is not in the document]
kkrempl: Yes, having the filepath in the metadata is probably a good idea!

src/ixdat/techniques/deconvolution.py
kkrempl: Yes, that's correct! I like you're suggestion with Kernel inheriting from Measurement, I added that as a TODO. For now I'll just leave it like it is to get the repo for the paper up and running.

src/ixdat/techniques/deconvolution.py
```diff
@@ -0,0 +1,254 @@
...
+        kernel_obj,
+        tspan=None,
+        t_bg=None,
+        snr=10):
+        """Returns the deconvoluted partial current for a given signal
+
+        Args:
+            signal_name (str): Name of signal for which deconvolution is to
+                be carried out.
+            kernel_obj (Kernel): Kernel object which contains the mass transport
```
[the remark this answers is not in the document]
kkrempl: Fully agree, I think this would be nice to have for the new Kernel object inheriting from Measurement

src/ixdat/techniques/deconvolution.py
kkrempl: Agree, on the TODO list for the new Kernel class (this was just copied pasted from the EC_MS deconvolution module).

src/ixdat/techniques/deconvolution.py
```diff
@@ -0,0 +1,254 @@
...
+
+    def gen_kernel(self,
+        dt=0.1,
+        len=100,
+        norm=True,
+        matrix=False
+        ):
+        """Generates a kernel/impulse response based on the defined mass
+        transport parameters contained in dict.
+
```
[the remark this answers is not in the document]
kkrempl: Agree!

src/ixdat/techniques/deconvolution.py
```diff
@@ -66,43 +64,40 @@ def extract_kernel(self, signal_name, cutoff_pot=0, tspan=None, t_bg=None):
...
-
 class Kernel:
-    """
-    Kernel class implementing different functionalities for datatreatment of
-    kernel/impulse response data.
-    """
-    def __init__ (
+    """Kernel class implementing datatreatment of kernel/impulse response data."""
+
+    # TODO: Make class inherit from Measurement, add properties to store kernel
```
ScottSoren, line 95: This is the only TODO I can find about kernel. I'm worried we'll forget the other ideas if this is all.
TODO's work well as multi-line bits of class docstrings. (PyCharm highlights them yellow 😄 )

src/ixdat/techniques/deconvolution.py
ScottSoren: Can you add a TODO here as well about storing these things?
Sorry to be annoying, but I don't think we'll look at these PR comments again so if it's not in the code it's nowhere.

src/ixdat/techniques/deconvolution.py
ScottSoren: Great. I can't find the TODO though.

ScottSoren (commented):
Hey, looks great! 
There's just a few more. One code change and a few more requests for TODO comments in the code. I actually couldn't find some of the TODO's you referred to... so I wonder if there's a file that didin't commit (or if I'm somehow looking at it a commit behind)?

src/ixdat/readers/ec_ms.py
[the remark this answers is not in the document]
ScottSoren: I guess there's nothing wrong with it. But I don't think it saves code compared to a dictionary. You can get col_list afterwards with `list(tseries_dict.keys())`

ScottSoren:
This PR does not make sense any more, as the master branch has changed a lot since it was initiated. I think the comments are still relevant for the deconvolution module. I have put a link to this PR in the module docstringand will close it.

PR #5, opened 2021-01-26
Debug tools, backend, and Measurement.__add__

ScottSoren:
Hi @KennethNielsen and @kkrempl , 

I want to merge my progress on this branch ASAP, because I caught a bunch of bugs before I could really start working on `ECMeasurement` and `CyclicVoltammagram`. With the changes here, the directory backend actually works 🤸 

I think it should not have any conflicts with #4 , so we can merge in either order, but this one should be merged before work on e.g. flushing out `ECMSMeasuremnt` for Zilien data begins.

### DIFF
--- a/development_scripts/ec_tools_development_pyCOox_data.py
+++ b/development_scripts/ec_tools_development_pyCOox_data.py
@@ -1,0 +1,47 @@
+# -*- coding: utf-8 -*-
+"""
+Created on Mon Jan 25 22:02:57 2021
+
+@author: scott
+"""
+from pathlib import Path
+from matplotlib import pyplot as plt
+
+from ixdat import Measurement
+
+plt.close("all")
+
+if False:
+    data_dir = Path(
+        "~/Dropbox/ixdat_resources/20B12_Data_Analysis_Workshop/example data set/"
+    )
+
+    ocp_file = data_dir / "01_Trimi1_cont_02_OCV_C01.mpt"
+    cv_file = data_dir / "01_Trimi1_cont_03_CVA_C01.mpt"
+    cp_file = data_dir / "01_Trimi1_cont_04_CP_C01.mpt"
+
+    ocp_meas = Measurement.read(ocp_file, reader="biologic", name="Pt_demo_ocp")
+    print("read ocp file!")
+    cv_meas = Measurement.read(cv_file, reader="biologic", name="Pt_demo_cv")
+    print("read cv file!")
+    cp_meas = Measurement.read(cp_file, reader="biologic", name="Pt_demo_cp")
+    print("read cp file!")
+
+    ocp_id = ocp_meas.save()
+    cv_id = cv_meas.save()
+    cp_id = cp_meas.save()
+else:
+    ocp_meas = Measurement.get(3)
+    cv_meas = Measurement.get(4)
+    cp_meas = Measurement.get(5)
+
+combined_meas = ocp_meas + cv_meas + cp_meas
+
+v_list = ["Ewe/V", "<I>/mA", "I/mA"]
+t_start = 13700
+tspan = [0, 5000]
+
+for meas in [ocp_meas, cv_meas, cp_meas, combined_meas]:
+    meas.tstamp += t_start
+
+    meas.plot(v_list=v_list, tspan=tspan)
--- a/development_scripts/ec_tools_development_pyCOox_data.py
+++ b/development_scripts/ec_tools_development_pyCOox_data.py
@@ -12,8 +12,8 @@ from ixdat import Measurement
 plt.close("all")
 
 if False:
-    data_dir = Path(
-        "~/Dropbox/ixdat_resources/20B12_Data_Analysis_Workshop/example data set/"
+    data_dir = Path.home() / (
+        "Dropbox/ixdat_resources/20B12_Data_Analysis_Workshop/example data set/"
     )
 
     ocp_file = data_dir / "01_Trimi1_cont_02_OCV_C01.mpt"
@@ -31,9 +31,9 @@ if False:
     cv_id = cv_meas.save()
     cp_id = cp_meas.save()
 else:
-    ocp_meas = Measurement.get(3)
-    cv_meas = Measurement.get(4)
-    cp_meas = Measurement.get(5)
+    ocp_meas = Measurement.get(7)
+    cv_meas = Measurement.get(8)
+    cp_meas = Measurement.get(9)
 
 combined_meas = ocp_meas + cv_meas + cp_meas
 
--- a/src/ixdat/backends/directory_backend.py
+++ b/src/ixdat/backends/directory_backend.py
@@ -135,6 +135,7 @@ class DirBackend(BackendBase):
         if not folder.exists():
             folder.mkdir()
         i = self.get_next_available_id(table_name)
+        obj_as_dict.update({"id": i})
         fixed_name = fix_name_for_saving(obj_as_dict["name"])
         file_name = f"{i}_{fixed_name}{self.metadata_suffix}"
 
@@ -153,7 +154,6 @@ class DirBackend(BackendBase):
         """Return the Path to the file representing row i of the table `table_name`"""
         folder = self.project_directory / table_name
         for p in folder.iterdir():
-            print("checking ")
             if id_from_path(p) == i and p.suffix == self.metadata_suffix:
                 return p
         print(f"could not find row with id={i} in table '{table_name}'")
--- a/src/ixdat/backends/directory_backend.py
+++ b/src/ixdat/backends/directory_backend.py
@@ -4,11 +4,29 @@ from .memory_backend import BackendBase
 from ..config import CFG
 
 
+char_substitutions = {
+    "/": "_DIV_",  # slash (divided by)
+    "\\": "_BKSL_",  # backslash
+    ".": "_DOT_",  # decimal
+    "^": "_CFLX_",  # circumflex accent (raised-to-the)
+    "<": "_LTS_",  # less-than sign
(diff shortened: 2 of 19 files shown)


development_scripts/ec_tools_development_pyCOox_data.py
```diff
@@ -0,0 +1,47 @@
...
+    cv_meas = Measurement.read(cv_file, reader="biologic", name="Pt_demo_cv")
+    print("read cv file!")
+    cp_meas = Measurement.read(cp_file, reader="biologic", name="Pt_demo_cp")
+    print("read cp file!")
+
+    ocp_id = ocp_meas.save()
+    cv_id = cv_meas.save()
+    cp_id = cp_meas.save()
+else:
+    ocp_meas = Measurement.get(3)
```
KennethNielsen, line 34: It may not matter, but I guess these id numbers aren't portable

src/ixdat/backends/directory_backend.py
```diff
@@ -4,11 +4,29 @@
...
+    "/": "_DIV_",  # slash (divided by)
+    "\\": "_BKSL_",  # backslash
+    ".": "_DOT_",  # decimal
+    "^": "_CFLX_",  # circumflex accent (raised-to-the)
+    "<": "_LTS_",  # less-than sign
+    ">": "_GTS_",  # greater-than sign
+}
+
+
+def fix_name_for_saving(name):
```
KennethNielsen, line 17: Good idea. Problems like these are annoying, because ultimately it can be difficult to make sure you account for everything. If it turns out to be a problem in the future, you can also take an approach where after replacing the most obvious, you simple try and create the file in a tmp dir (https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory) and see if it succeeds. Then if it fails, replace the offending characters with blank spaces or the Unicode code point name.

src/ixdat/backends/directory_backend.py
```diff
@@ -27,24 +45,33 @@ class DirBackend(BackendBase):
...
+            directory (Path): the main ixdat directory
+            project_name (str): the name of the project (ixdat subdirectory)
             metadata_suffix (str): The suffix to use for JSON-formatted metadata files
             data_suffix (str): The suffix to use for numpy-formatted data files
         """
-        self.directory = directory
+        self.project_directory = directory / project_name
+        if not self.project_directory.exists():
+            try:
+                self.project_directory.mkdir()
```
KennethNielsen, line 63: I don't know if this path might reach into several layers of new directories. If so, you might want to add `parents=True`   https://docs.python.org/3/library/pathlib.html#pathlib.Path.mkdir (also says something about exceptions) and then, you might add `exists=True` and remove the if.

src/ixdat/config.py
```diff
@@ -15,9 +15,10 @@ class Config:
     """
 
     def __init__(self):
-        self.standard_metadata_suffix = "ix"
-        self.standard_data_suffix = "ixdata"
+        self.standard_metadata_suffix = ".ix"
```
KennethNielsen, line 18: Good. Adding the dot in there, also seems compatible with (https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.suffix)

src/ixdat/backends/directory_backend.py
```diff
@@ -4,11 +4,29 @@
...
+    "\\": "_BKSL_",  # backslash
+    ".": "_DOT_",  # decimal
+    "^": "_CFLX_",  # circumflex accent (raised-to-the)
+    "<": "_LTS_",  # less-than sign
+    ">": "_GTS_",  # greater-than sign
+}
+
+
+def fix_name_for_saving(name):
+    """Replace problematic characters in name with the substitution defined above"""
```
KennethNielsen, line 18: substitutionS

src/ixdat/measurements.py
```diff
@@ -68,21 +69,22 @@ def __init__(
...
+            sample (Sample): The (already loaded) sample being measured
+            sample_name (str): The name of the sample being measured (will be loaded)
             lablog (LabLog): The log entry with e.g. notes taken during the measurement
             tstamp (float): The nominal starting time of the measurement, used for
                 data selection, visualization, and exporting.
         """
         super().__init__()
         self.name = name
         self.technique = technique
-        if isinstance(metadata, str):
```
KennethNielsen, line 79: I think I actually kind of preferred the single argument multiple types form, simply because the ambiguity of what happens if you provide both disappear. But I must admit that I can come up with arguments both ways. What was your reason for changing it?

Another question for this is (I may have asked this before), will the json-string form ever be used by anything else than classmethods, because if not, they should convert to normal dict since it is unambiguous in their body and then the dual mode could disappear.

KennethNielsen (changes requested):
Good. Nothing much to this.

In the interest of full disclosure this is what I did:
- I proofread the tools parts thoroughly
- I proofread as much of the local semantics as was visible in the diff for all the small changes
- I proofread the text parts of everything else

I unfortunately **did not** have time to proofread all the small fixes properly into context :(

src/ixdat/measurements.py
ScottSoren: It was an unresolved comment to #1 whether the responsibility of converting between dicts and json_strings (which at least the directory backend would need) belongs with the Saveable class or with the backend. 
https://github.com/ixdat/ixdat/pull/1#discussion_r546436991
I just saw your last comment there now (despite effort, I'm not getting the github notifications I want, must look into that again). Okay great, I am persuaded. `Measurement.metadata` will only exist as a dictionary. The directory backend will dump this to / load from a json string before saving/initiating. Other backends may do whatever.

Relevant, though, elsewhere in `ixdat`, is the one vs two arguments discussion:
The reason for two arguments is so that the type of any one of them is never ambiguous - I think a comment of yours in an earlier review encouraged me to try that (though maybe not right here).
Some clumsiness in these `.__init__`s in general I think has to do with the `Saveable` framework for class-table correspondence. The columns of tables now have to have the same names as the attributes that go into them.

src/ixdat/config.py
ScottSoren: Yeah, was scratching my head about files ending with `..ix` (two dots).
Also it seems numpy.save cannot will always add `.npy` extension if it's not there.

development_scripts/ec_tools_development_pyCOox_data.py
ScottSoren: Right. But they fit in human short-term memory :D

src/ixdat/backends/directory_backend.py
ScottSoren: This is enough for all BioLogic columns so far. TODO left with link, for when we need the solution to be guaranteed for new evil datasets.

src/ixdat/backends/directory_backend.py
ScottSoren: fixed.

src/ixdat/backends/directory_backend.py
ScottSoren: Thanks for the tricks 😄 . I've made the improvement here, and also in config.py.

src/ixdat/measurements.py
ScottSoren: Done. Actually, `DirectoryBackend` simply dumps the `Saveable.as_dict` into json, and json has no problem with nested dictionaries, so `metadata` does not need to be converted to string at all for this backend.
Also, I left the `Measurement.metadata_json_string` property to remind anyone who sees it not to put unreadable stuff into `metadata`, but right now nothing uses the property.
It is not yet clear to me how another database type like mariaDB will know that it has to convert it to AND from json strings.
Also made `sample` one argument. But this is where it gets hairy, because what is stored in the db is `sample_name` whether `__init__()` is given `sample` as the Sample or as a string that it uses to load the sample. A similar problem is `DataSeries.unit` vs `DataSeries.unit_name`. One way (which I did first) to handle this is to make the Saveable class's `column_attrs` a dictionary rather than a set as it is now (see commit https://github.com/ixdat/ixdat/pull/5/commits/21017f40b9faadcb902fe60e387dece58148fbef for the conversion). But this is really clumsy and verbose, since most attributes are 1:1 with their column names.
I just implemented a new solution in `Measurement.from_dict()` which I agree is a reasonable place to handle this kind of thing. I've left a long comment there as well describing the problem and linking to this github comment.

Please resolve this and approve the PR if you think this is okay for now.

kkrempl (approved):

src/ixdat/measurements.py
KennethNielsen: I think this is a design issue that may warrant a little more discussion, but by all means let's move forward now.

As for both the MariaDB question and the `sample` problem I think that we need to go back to where `column_attrs` are a little more complex. So one way that I could imagine solving this is give a type for each column in `column_attrs` it could be trivial `int, bool, float` or something more complex like `json` but it could also be something custom, like `sample`. Each of the backends should then know, how to create the proper columns from all those types, including the custom ones, and how to convert back and forth.

development_scripts/ec_tools_development_pyCOox_data.py
KennethNielsen: Oh. I meant more that I though that the fact that the turned out to be number 3, 4 and 5 is an artifact of the time at which you ran it and that the test script might need to store those ids somehow, for someone else to be able to run the script. But as I said, it may be a bit of a stretch for now.

KennethNielsen (approved):

src/ixdat/measurements.py
ScottSoren: okay. Looking forward to more discussion!

PR #6, opened 2021-01-28
basic EC features

ScottSoren:
Hey @KennethNielsen and @kkrempl ,

Here I've flushed out the ECMeasurement class and made a CyclicVoltammagram class that adds a few features. The main focus of this PR is on how to make sure the essential data series (potential, current, and selector in this case, see class docstring to `ECMeasurement`) are always available by a standard name, even though they might be saved differently in different files or not present in all appended files. It proved slightly hair-pulling, even with the relative strengths of `ixdat`'s structure.
The "front-panel" (user-facing) scripts I've been developing this with are `development_scripts/ec_tools_dev.py` and `development_scripts/cv_tools_dev.py`. They show how easy it is to manipulate and visualize a complex looped experiment with OCP, CVA and CP (ask me for the data, shared some of ec_tools_dev.py's plots in Slack yesterday).
The inside is a bit of a mess - I've used slightly different approaches to build and cache different variables, when it would be much better to have a general approach. One of @KennethNielsen 's ideas (see Slack) was to make it the Reader's responsibility to rename to the standard names and make place-holders for missing essential data. My instinct is that it is more natural to give the Technique Measurement class that responsibility, and only require the Reader to provide static lists of what things are called in the files it reads, but hope we can find the best solution while discussing the code.
It would also be nice to move a lot of the code that solves general problems to a `build` module or something to reduce the clutter in the technique class.
There are TODO's and FIXME's all over the code with problems and (suboptimal) ideas to improve, so especially loook at those. Hope you we can find better solutions to these problems!
Thanks!

### DIFF
--- a/src/ixdat/measurements.py
+++ b/src/ixdat/measurements.py
@@ -51,8 +51,6 @@ class Measurement(Saveable):
 
         Args:
             name (str): The name of the measurement
-            TODO: Decide if metadata needs the json string option.
-                See: https://github.com/ixdat/ixdat/pull/1#discussion_r546436991
             metadata (dict): Free-form measurement metadata. Must be json-compatible.
             technique (str): The measurement technique
             s_ids (list of int): The id's of the measurement's DataSeries, if
@@ -89,6 +87,7 @@ class Measurement(Saveable):
             component_measurements, m_ids, cls=Measurement
         )
         self.tstamp = tstamp
+        self.sel_str = None  # the default thing to select on.
 
     @classmethod
     def from_dict(cls, obj_as_dict):
@@ -204,12 +203,12 @@ class Measurement(Saveable):
     @property
     def series_names(self):
         """List of the names of the series in the measurement"""
-        return [series.name for series in self.series_list]
+        return set([series.name for series in self.series_list])
 
     @property
     def value_names(self):
         """List of the names of the VSeries in the measurement's DataSeries"""
-        return [vseries.name for vseries in self.value_series]
+        return set([vseries.name for vseries in self.value_series])
 
     @property
     def value_series(self):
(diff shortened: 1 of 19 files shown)


src/ixdat/measurements.py
```diff
@@ -269,22 +286,203 @@ def data_cols(self):
...
+        for series in self.series_list:
+            try:
+                tseries = series.tseries
+                if tseries is None:
+                    raise AttributeError
+            except AttributeError:  # series independent of time are uneffected by cut
+                new_series_list.append(series)
+            else:
+                t_id = (tseries.id, tseries.backend_name)
+                # FIXME: Beautiful, met my first id clash here. Local memory and loaded
```
kkrempl, line 347: Would a solution to this be using large random numbers as identifiers?

src/ixdat/measurements.py
```diff
@@ -269,22 +286,203 @@ def data_cols(self):
...
+            for value in allowed_values:
+                m = new_measurement.select_value(**{series_name: value})
+                if meas:
+                    meas = meas + m
+                else:
+                    meas = m
+            new_measurement = meas
+        return new_measurement
+
+    def select(self, *args, tspan=None, **kwargs):
```
kkrempl, line 477: Not sure if I understand correctly what the select_x functions exactly do. Is it to return a measurement object that contains e.g. only the signal for M2 in a given tspan?

src/ixdat/measurements.py
```diff
@@ -333,49 +531,74 @@ def __add__(self, other):
 
 
 #  ------- Now come a few module-level functions for series manipulation ---------
```
kkrempl, line 533: Is this where we would also have integration functions? (such as in Kernel.sig_area?)

kkrempl (commented):
Just a couple questions for the measurement class :)

src/ixdat/techniques/ec.py
```diff
@@ -1,18 +1,597 @@
...
+        E_str="raw potential / [V]",
+        V_str="$U_{RHE}$ / [V]",
+        RE_vs_RHE=None,
+        R_Ohm=None,
+        raw_potential_names=("Ewe/V", "<Ewe>/V"),  # TODO: reader must define this
+        I_str="raw current / [mA]",
+        J_str="J / [mA cm$^{-2}$]",
+        A_el=None,
+        raw_current_names=("I/mA", "<I>/mA"),  # TODO: reader must define this
+        cycle_names=("cycle number",),
```
kkrempl, line 115: One comma too much after ")"

src/ixdat/techniques/ec.py
```diff
@@ -1,18 +1,597 @@
...
+            "A_el",
+            "raw_current_names",
+        }
+    }
+
+    def __init__(
+        self,
+        name,
+        *,
+        technique=None,
```
kkrempl, line 92: Isn't it inheriting some of these kwargs already from the __init__ of the Measurement class? Is there a reason to have them here again?

src/ixdat/measurements.py
ScottSoren: Yeah, Kenneth has made the same point. I first resonded that I want to see it become necessary first (see https://github.com/ixdat/ixdat/pull/1#discussion_r546400226 and https://github.com/ixdat/ixdat/pull/1#issuecomment-754203224). And here it is. But now I have a new argument which is using [UUID](https://en.wikipedia.org/wiki/Universally_unique_identifier) instead lose the readability of the id's, which I find very useful when playing with it and the `DirBackend` (see slides). 
It could be good to have both the small-integer id and a UUID. 
But this (id, backend_name) could also work for a while. I think in any case it can wait til we have an actual SQL backend to test things against. There's TODO's in the code so we won#t forget ;)

src/ixdat/measurements.py
ScottSoren: No, the measurement it returns contains every `DataSeries` which has data corresponding to the time being selected for! Select just gives access to cutting and selecting in the same function.
You can say `select(tspan=[100, 200], file_number=1)`, for example.
Ideally you should be able to say e.g., `select(cycle=1, 0.5<potential<1)`, but that's not implemented yet.

Is this clear? Does something need to be added to the docstring or changed?

src/ixdat/measurements.py
ScottSoren: Not here. Those would be bound to a class.

But I can't decide whether it should go in `Measurement` or in `ValueSeries`. In principle the latter has all it needs (the value np array and the time np array). Which do you prefer of these two (`meas` is a `Measurement` instance)?
1. `int_M2 = meas.calc_integral("M2", tspan=[100, 200])`
2. `int_M2 = meas["M2"].calc_integral(tspan=[100, 200])`

src/ixdat/techniques/ec.py
```diff
@@ -1,18 +1,597 @@
...
+            self[
+                self.I_str
+            ] = self._raw_current  # TODO: better cache'ing. This is saved
+        else:
+            raise SeriesNotFoundError(
+                f"{self} does not have a series corresponding to raw current."
+                f" Looked for series with names in {self.raw_current_names}"
+            )
+
+    def calibrate(self, RE_vs_RHE, A_el=None):
```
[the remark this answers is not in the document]
ScottSoren: Oh, fair point. Done. Renamed `calibrate_RE`. 
An argument could be made to not having it there at all, and letting the user just set it directly, e.g., `meas.RE_vs_RHE = 0.715`
I left a `calibrate` though which can set `RE_vs_RHE`, `A_el`, and/or `R_Ohm` in one go. This should I guess be overwritten by e.g. `ECMSMeasurement`.

src/ixdat/techniques/ec.py
```diff
@@ -1,18 +1,597 @@
...
+        """Calibrate the reverence electrode by providing `RE_vs_RHE` in [V]."""
+        self.RE_vs_RHE = RE_vs_RHE if RE_vs_RHE is not None else self.RE_vs_RHE
+        if A_el:  # So that you can calibrate and normalize in one method call:
+            self.normalize(A_el=A_el)
+
+    def correct_ohmic_drop(self, R_Ohm=None):
+        """Correct for ohmic drop by providing `R_Ohm` in [Ohm]."""
+        self.R_Ohm = R_Ohm if R_Ohm is not None else self.R_Ohm
+
+    def normalize(self, A_el):
```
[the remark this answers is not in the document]
ScottSoren: Done.

src/ixdat/techniques/ec.py
ScottSoren: I wrote it all here to be explicit, so that `help(ECMeasurement)` will give all the info. I don't know how else to ensure that. Perhaps @KennethNielsen does.

ScottSoren:
This has been superceded by #11 . Closing.

PR #9, opened 2021-03-10
CyclicVoltammatry and ECMS

ScottSoren:
Hey @kkrempl , this PR contains implementation to your review of #8. It also contains new readers for two other potentiostat files as well as PVMassSpec, which would be cool for you to take a look at if you have time. An idea could be to leave this pull request open as we do a bit of work on ECMSMeasurement.

One thing I'd like a second opinion on is the way MID ValueSeries are made accessible as e.g. "M2", even if the data acquisition software saves them as e.g. "2_amu" instead. That's done, like probably too many things, by adding to `__getitem__` . I worry that complicates inheritance. I am however reluctant to change the names of the ValueSeries since I think it's a nice property of ixdat if it respects the original names even while adding accessibility via standard names (same storry for potential and current in ECMeasurement). Another way to do this would be to make a `ReferenceSeries` which is like a `DataSeries` but doesn't have a copy of the data but just the standard name and a pointer towards the `DataSeries` with the original name and the data. What do you think?


ScottSoren:
I'll push now the most recent changes to this branch now as well, so you can use the PR to give a look at all the new code in the present pip-installable v0.0.10
I'll refrain from any edits except maybe documentation for a bit so it stays current.

kkrempl (commented):
Hi, I finally went through this version of the code and tried it out on some of my data. It worked well and I could get the correct EC-MS as well as EC-MS-CV plots out of it! :) I like the interface and I think in general we should adhere to the principle that, if not explicitly defined, the grab_xxxx commands return the data w/o any modifications.

ScottSoren:
@kkrempl, glad to hear!
With regards to `grab`, did you experience anywhere where it didn't do the desired behaviour of returning unmodified data by default?

kkrempl:
> @kkrempl, glad to hear!
> With regards to `grab`, did you experience anywhere where it didn't do the desired behaviour of returning unmodified data by default?
No, with the change to include_endpoints=False I have not had any problems so far.

PR #18, opened 2021-11-22
Test restructure

KennethNielsen:
This is the first part of a restructuring/refactoring of the existing test and a little expansion of the test suite.

## Test file structure

The existing test suite consisted largely of functional tests on an ECMeasurement and these functional tests on this object itself was combined with round trip tests through directory backend save/load and through `as_dict`/`from_dict` to test that everything in the ECMeasurement survived the trip. While this first iteration of tests was very successful in aiding during development, the structure made is difficult to figure out exactly what was tested where and if some things were being tested more than once. Tests, also functional tests, should ideally test just one thing, so the first step of the refactoring was to brake these tests up into two modules. This was combined with a new test folder structure, which lends itself towards lots of tests of different kinds. The results structure is:
```text
tests/
├── functional
│   ├── conftest.py
│   ├── __init__.py
│   ├── test_backends.py
│   └── test_measurements.py
└── README.md
```
where conftest.py contains some shared test fixtures (more about those below). The structure can then be expanded e.g. with unit tests in a folder called "unit" at the same level as "functional"

## Equality

As mentioned, a central part of the rewrite was to separate the two purposes mentioned above; ECMeasurement functionality (of which testing very little was changed) and round trip survival. The latter leads into a discussion about how exactly to test for whether an object survives the trip. In the ideal, I think, what we should check for is whether an object is **equal** to whatever arrives in the other end, i.e:
```python
id_ = measurement.save()
reloaded_measurement = Measurement.get(id_)
assert measurement == reloaded_measurement
```
Then we need to figure out, how exactly do we define equality for our objects. A couple of interesting questions came up:
1. Is the backend, and thereby all the ids, required to be equal?
2. What about the data, is it required to be equal down to the last bit, in order to be equal?

The decisions I landed on, and which obviously are open to discussion, were:
1) No. For all of the types we have, that can be saved and loaded, I consider this saving and loading something orthogonal to the object itself, and I therefore do **not** consider ids of objects or the backend when comparing equality. This has as a consequence that we **do** have to test equality on the data itself
2) I personally think it is too strict a requirement on a backend, and other serialization, that the data has to arrive back unaltered down to the last bit. I therefore opted to test for data equality with `numpy.allclose` which tests `rtol=1e-05, atol=1e-08`. Obviously the thresholds could be modified by making a local [partial](https://docs.python.org/3/library/functools.html#partial-objects) in tools or utils.

With these two decisions in mind, the `__eq__` method was implemented for all the necessary objects and the tests merely written as above. A side note to implementing `__eq__` for a custom object is that doing so, sets the implementation of `__hash__` to None (see here https://docs.python.org/3/reference/datamodel.html#object.__hash__). This is problem, because the trick of passing a sequence of objects through `set` to remove duplicates is used several places, but this requires the object to be hashable. This is quickly fixed by borrowing the __hash__ function from `object` (which I think essentially just returns the memory address). This has the implication that only the same object will have the same hash. As mentioned in the docs above, this may be a problem, because the implementation or `__eq__` are "expected to be" related to the implementation of `__hash__`. Ours are not and with the decision on 2 probably never will. This has the amusing and somewhat exotic effect that trying to combine series which are the same, but with different backends, will **not** remove the "duplicates" from the series list, even though the object is equal. This may not be the intention, but that is also the way it already was.

One unresolved problem that I ran into, is that Calibration has an optional back reference to the measurement it should work on by default. Since I would normally include all objects in the equality check, this would result in a circular reference and infinite recursion. For now this is "solved" by simply **not** including this measurement in the Calibration equality check.

The decision on 2, in that I meant it to also apply to metadata, meant that some helper functions was written in `tools.py` to compare equality on dicts and lists, but where floats in the data structure are only required to be close (using `np.isclose`).

## Tests improvements

A few but important improvements was done to make use of some of the testing framework:

- Setup of dependencies that is used in multiple tests, was factored out of the test code by means of [fixtures](https://docs.pytest.org/en/6.2.x/fixture.html)
- Testing load/save to the directory backend was altered so that every test happens on a completely fresh directory backend in a tmp dir. This also happens by the magic of fixtures. I also fixed a bug found during this testing. 
- Several tests on float values was changed to use [pytest approx](https://docs.pytest.org/en/6.2.x/reference.html#pytest-approx) to make sure they will keep working even if more manipulation is done on them, which might have float rounding problems.

## Future work and proof reading focus

I think a natural next step is to remove the last duplication of tests that still exists, and say that if we really trusts the round trip tests, then it should not be necessary to do functionality tests across round trips. This of course means that it is super important that we all trust and completely agree on the round trip tests that depends on `__eq__` so those should get at much attention as possible.

I hope you like it :)


ScottSoren:
Hi Kenneth,

Awesome philosophical stuff. Agree with the test restructure and the need for an `__eq__`, and all your considerations about what `__eq__` should mean.

I think `__eq__`  should make use of the `as_dict()` representation and the class attribute `child_attrs`. If two `Saveable` objects have the same keys in their dictionary representations (`self_as_dict` and `other_as_dict`) AND for every `key` either (i) each object pointed to in  the lists `self_as_dict[key]` and  `other_as_dict[key]` are equal if `key` referes to a "child attr" and (ii) the two values are equal (or `np.allclose` for numerical data) otherwise.
I think that then `__eq__` (and `__hash__` need only be implemented **once** and away from the eyes of users or most developers, in `db.Saveable`.

I want to write test this, however, I am stuck with how to work on this and need a bit of help: 

1. You're making this PR on a fork rather than a branch, so don't see how to add commits to the PR it in a straightforward way. Should I (i) approve this and make my own PR with what I figure out 

2. I can't run the tests :( ... this is a me problem, but indicates that the barrier to entry is still quite high to all the stuff in **tools.rst**. I've tried the following (attempts i and ii give the same result in git bash):

  (i) `tox` in the Anaconda Powershell Prompt. I get 
  ModuleNotFoundError: No module named 'distlib'
  and 
  ERROR:   py37: InvocationError for command 'c:\users\scott\anaconda3\python.exe' -m virtualenv --no-download --python 'c:\users\scott\anaconda3\python.exe' py37 (exited with code 1)
  
  (ii) `pytest` in Anaconda Powershell Prompt. I get 
  ERROR  - ModuleNotFoundError: No module named 'ixdat'
  
  (iii) Doing it manually. Running a module **rests/functional/run_test.py** with the content
  ```
  from .test_measurements import TestBiologicEC
  from .conftest import ec_measurement as get_ec_measurement
  
  ec_measurement = get_ec_measurement()
  
  tester = TestBiologicEC()
  
  tester.test_calibrate_and_append(ec_measurement)
  ```
  gives the error 
  ModuleNotFoundError: No module named '__main__.test_measurements'; '__main__' is not a package

I only had a hour to try and figure out how to run tests today. DIdn't find a clear answer in **tools.rst** which I think means  that documentation is incomplete. Didn't start googling for answers. Suggestions welcome!

tests/functional/test_backends.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+class TestBackends:
+    """Functional tests for the data backends"""
+
+    def test_round_trip(self, ec_measurement, fresh_directory_backend):
+        """Test load/save round trip of a measurement"""
+        id_ = ec_measurement.save()
+        reloaded_ec_measurement = Measurement.get(id_)
+        assert ec_measurement == reloaded_ec_measurement
```
ScottSoren, line 23: Nice with the use of `__eq__` !

tests/functional/test_measurements.py
```diff
@@ -0,0 +1,96 @@
...
+
+        # Finally, save and load the copied measurement and check that it's still there:
+        # FIXME the load save test add to the total test time, this one is a primary
+        # candidate to be removed if we properly trust the backend tests
+        i2 = meas12_copied.save()
+        meas12_copied_loaded = Measurement.get(i2)
+        assert meas12_copied_loaded.RE_vs_RHE == RE_vs_RHE
+        assert len(meas12_copied_loaded.v) == len(composed_measurement.v)
+
+    def test_to_and_as_dict(self, composed_measurement):
```
ScottSoren, line 93: "to_and_from" would be better English. "as_and_from" would mirror the method names `as_dict()` and `from_dict()`

tests/functional/test_measurements.py
```diff
@@ -0,0 +1,96 @@
...
+        assert cvs_1_plus_2.RE_vs_RHE == ec_measurement.RE_vs_RHE
+        # Check that the main time variable, corresponding to potential, wasn't corrupted:
+        assert len(cvs_1_plus_2.grab("potential")[0]) == len(
+            cvs_1_plus_2["time/s"].data
+        )
+        # Check that the selector is still available and works with the main time variable:
+        assert len(cvs_1_plus_2.selector.data) == len(cvs_1_plus_2.t)
+
+    def test_calibration_over_save_load(self, composed_measurement):
+        """Test save and load functionality"""
```
ScottSoren, line 46: This function looks like it could be broken up more.

src/ixdat/data_series.py
```diff
@@ -45,6 +45,21 @@ def from_dict(cls, obj_as_dict):
     def __repr__(self):
         return f"{self.__class__.__name__}(id={self.id}, name='{self.name}')"
 
+    def __eq__(self, other):
+        """Return whether this DataSeries is equal to another"""
```
ScottSoren, line 49: TODO with "metaprogramming": Only one implementation, `Saveable.__eq__`

src/ixdat/data_series.py
```diff
@@ -90,6 +105,18 @@ def tseries(self):
         """Trivially, a TimeSeries is its own TimeSeries"""
         return self
 
+    def __eq__(self, other):
+        """Return whether this TimeSeries is equal to another"""
```
ScottSoren, line 109: TODO with "metaprogramming": Only one implementation, `Saveable.__eq__`

src/ixdat/data_series.py
```diff
@@ -178,6 +205,21 @@ def data(self):
                 )
         return self._data
 
+    def __eq__(self, other):
+        """Returns whether this Field is equal to another"""
```
ScottSoren, line 209: TODO with "metaprogramming": Only one implementation, `Saveable.__eq__`

src/ixdat/measurements.py
```diff
@@ -882,6 +884,42 @@ def __add__(self, other):
         del obj_as_dict["s_ids"]
         return cls.from_dict(obj_as_dict)
 
+    def __eq__(self, other):
+        """Return whether this object is equal to `other`"""
```
ScottSoren, line 888: TODO with "metaprogramming": Only one implementation, `Saveable.__eq__`

src/ixdat/measurements.py
```diff
@@ -908,6 +946,14 @@ def __init__(self, name=None, technique=None, tstamp=None, measurement=None):
         self.tstamp = tstamp or (measurement.tstamp if measurement else None)
         self.measurement = measurement
 
+    def __eq__(self, _):
+        """Return whether the calibration is equal to another
```
ScottSoren, line 950: TODO with "metaprogramming": Only one implementation, `Saveable.__eq__`

ScottSoren (approved):
Great! 
I got pytest to work fine, just had to `pip install -e ixdat` in my clone of this fork.
With that, I'm able to run the tests, and I approve! I will add a note of that in **tools.rst**.

(I'm still having a problem with tox for my clone of this fork, but it's regardless of which branch I look at, whereas tox works fine on the master branch of my clone of the main fork at ixdat/ixdat. So it's something about a broken venv from pycharm or anything, nothing to do with the code in this PR)

I've put some comments here, tiny stuff. The only big thing is that __eq__ should only be defined in my opinion once, in the parent class `db.Saveable` . But that's definitely a "metaprogramming" thing, so I will make an implementation in the metaprogramming branch.

To get that started now without creating a complex merge, I will approve this PR now and get it merged :) 

Woohoo! ixdat has got a good test structure. 🤸

ScottSoren:
Here is the `Saveable.__eq__` mentioned above: https://github.com/ixdat/ixdat/blob/c050207a494df12a404dfbd75d1cddc7f7a09de9/src/ixdat/db.py#L308

PR #28, opened 2021-12-02
One __eq__

ScottSoren:
As discussed, __eq__ is generalized to be in Saveable.

A full metaprogramming PR with table defs and improved serialization structure will come later.


KennethNielsen (approved):
Approve with small changes

PR #34, opened 2022-03-02
[RFC] Deprecation decorator

KennethNielsen:
This is a Request For Comments on the idea of introducing a decorators that would mark either a class, function methods **or** a specific kwarg within any of these callables as deprecated. After input I will make a more polished PR. What I need input on is this:

1. The implementation has both the notion of soft and hard deprecation, where the former issues a warning and the latter raises and exception. Personally, I think that is nice, but I need input on whether you think it is overkill
2. The periods where a feature is soft or hard deprecated in given as a time interval since last supported release. I.e "the feature is soft deprecated for all releases for 3 months after version '1.2.3' and hard deprecated an additional 3 months after that". Is this the way you prefer to work with these intervals? One might also list those intervals as a number of releases i.e. "the feature is soft deprecated 2 releases after version '1.2.3' and hard deprecated an additional 2 releases after that"
3. EDIT How clever do we want it to be? In principle we could of course give it a "last supported time stamp" and well defined time deltas and then we could calculate when it would move from soft deprecation to hard, as opposed to the suggested implementation when you will need to flip the `hard_deprecation` boolean by hand. I didn't do it because it is my view that it will not be worth the extra code complexity and because I don't want peoples code to all of a sudden hard deprecate with them having upgraded.

The feature is used like this:
```python
import sys
from time import sleep

from ixdat.tools import deprecate


@deprecate("1.2.3", "1 month", "1 month", "Please use `MyNewClass` instead")
class MyClass:
    pass

my_obj = MyClass()

# This is just a bit of trickery to make the warnings appear in order with the printed
# separator
sys.stderr.flush()
sleep(0.1)
print("\n\n===\n\n", flush=True)

class MyClass:

    @deprecate("1.2.3", "1 month", "1 month", "Please use `mynewmethod` instead")
    def mymethod(self):
        ...

    @classmethod
    @deprecate(
        "1.2.3",
        "1 month",
        "1 month",
        "Please use `kwargb` instead",
        kwarg_name="kwarga",
        hard_deprecate=True
    )
    def myothermethod(cls, arg, kwarga=None, kwargb=None):
        ...


my_obj = MyClass()
my_obj.mymethod()  # Will warn

sys.stderr.flush()
sleep(0.3)
print("\n\n===\n\n", flush=True)
sleep(0.3)


my_obj.myothermethod(1, kwarga=7)
```

which outputs something like this:
```python
/home/kenneth/venv/ixdat/ixdat/test.py:12: DeprecationWarning: The class 'MyClass' is deprecated, its last supported version being 1.2.3.
This means that:
* It is soft deprecated (issuing warnings) in all releases for 1 month after the last supported release
* It is hard deprecated (raising exceptions) in all releases for an additional 1 month after that
* After the hard deprecation period ends, it will be removed

See instruction below on how to update your code to avoid this message:
Please use `MyNewClass` instead
  my_obj = MyClass()


===


/home/kenneth/venv/ixdat/ixdat/test.py:40: DeprecationWarning: The method 'MyClass.mymethod' is deprecated, its last supported version being 1.2.3.
This means that:
* It is soft deprecated (issuing warnings) in all releases for 1 month after the last supported release
* It is hard deprecated (raising exceptions) in all releases for an additional 1 month after that
* After the hard deprecation period ends, it will be removed

See instruction below on how to update your code to avoid this message:
Please use `mynewmethod` instead
  my_obj.mymethod()  # Will warn


===


Traceback (most recent call last):
  File "/home/kenneth/venv/ixdat/ixdat/test.py", line 48, in <module>
    my_obj.myothermethod(1, kwarga=7)
  File "/home/kenneth/venv/ixdat/ixdat/src/ixdat/tools.py", line 211, in inner_function
    raise DeprecationError(compound_message)
ixdat.exceptions.DeprecationError: The property named 'kwarga' in method 'MyClass.myothermethod' is deprecated, its last supported version being 1.2.3.
This means that:
* It is soft deprecated (issuing warnings) in all releases for 1 month after the last supported release
* It is hard deprecated (raising exceptions) in all releases for an additional 1 month after that
* After the hard deprecation period ends, it will be removed

See instruction below on how to update your code to avoid this message:
Please use `kwargb` instead
```

### DIFF
--- a/src/ixdat/exceptions.py
+++ b/src/ixdat/exceptions.py
@@ -31,3 +31,7 @@ class TechniqueError(Exception):
 
 class QuantificationError(Exception):
     """ixdat errors having to do with techniques and their limitations"""
+
+
+class DeprecationError(Exception):
+    """Error raised when a deprecated component has reached hard deprecation"""
--- a/src/ixdat/tools.py
+++ b/src/ixdat/tools.py
@@ -1,0 +1,222 @@
+"""This module contains general purpose tools"""
+import inspect
+import warnings
+from functools import wraps
+
+import numpy as np
+
+from ixdat.exceptions import DeprecationError
+
+warnings.simplefilter("default")
+
+
+def thing_is_close(thing_one, thing_two):
+    """Return whether two things are (nearly) equal, looking recursively if necessary"""
+    if type(thing_one) != type(thing_two):
+        return False
+
+    if isinstance(thing_one, list):
+        return list_is_close(thing_one, thing_two)
+    elif isinstance(thing_one, dict):
+        return dict_is_close(thing_one, thing_two)
+    else:
+        return value_is_close(thing_one, thing_two)
+
+
+def value_is_close(value_one, value_two):
+    """Return whether `value_one` and `value_two` are equal (or close for floats)"""
+    if isinstance(value_one, float) or isinstance(value_two, float):
+        return np.isclose(value_one, value_two)
+    elif isinstance(value_one, np.ndarray) and isinstance(value_two, np.ndarray):
+        return np.allclose(value_one, value_two)
+
+    return value_one == value_two
+
+
+def dict_is_close(dict_one, dict_two):
+    """Return True if the dicts are equal, except floats are allowed to just be close
+
+    Args:
+        dict_one (dict): The first dictionary to compare
+        dict_two (dict): The second dictionary to compare
+
+    .. warning::
+       This function is recursive (also together with :ref:func:`list_is_close` and
+       **will** result in an infinite loop if the values reference back to itself
+    """
+    if len(dict_one) != len(dict_two):
+        return False
+    if dict_one.keys() != dict_two.keys():
+        return False
+
+    for key, value_one in dict_one.items():
+        value_two = dict_two[key]
+        if not thing_is_close(value_one, value_two):
+            return False
+
+    return True
+
+
+def list_is_close(list_one, list_two):
+    """Return True if the lists are equal, except floats are allowed to just be close
+
+    Args:
+        list_one (list): The first list to compare
+        list_two (list): The second list to compare
+
+    .. warning::
+       This function is recursive (also together with :ref:func:`list_is_close` and
+       **will** result in an infinite loop if the values reference back to itself
+    """
+    if len(list_one) != len(list_two):
+        return False
+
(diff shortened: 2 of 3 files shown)


ScottSoren:
Hi Kenneth, 

I think all of this is awesome! The useage syntax is simple and easy enough that I would indeed use it all over the place!
The only thing I think is overkill is the timed moving from soft to hard depreciation. I think it would be better to do that by versions. That gives us more flexibility, important since ixdat is not anyone's full job. If, say, we set up a depreciation in an ixdat release but then no one has time for ixdat for a few months and, in the meanwhile it is revealed that the thing should not have been depreciated, we'd have shot ourselves in the foot with the timer but have caused approx. zero damage with the versions.
So maybe the code could be
```
class MyClass:
    @deprecate("1.2.3", "1.3.0", "Please use `kwargb` instead", kwarg_name="kwarga", )
```
to soft depreciate use of kwarga after 1.2.3 and hard depreciate after 1.3.0... something like that?

I also like the `only_one_of` decorator, and agree that could clear up some complex kwarg interpreting at the start of a number of methods.

KennethNielsen:
@ScottSoren It's a good point about avoiding the timed hard deprecation. I will change it. Only thing is, if you provide a specific version number as opposed to say "3 minor/major releases after" then you are a little locked in with what the next versions will be. It might not be a problem in reality, what do you think?

KennethNielsen:
Ah. I think there is a small terminology difference here. I refer to the second digit as "minor", as per [semver](https://semver.org/). In any case, since you are primary maintainer, if you would prefer a specific version to be listed, I think we should do that.

Making the deprecation thresholds optional is a good idea. I will change the implementation to accommodate this. What this means is that:
* Applying the decorator means the feature is immediately (in practice only noticed from next release) soft deprecated
* The developer may optionally list at which versions the feature will be hard deprecated
* The developer may optionally list at which version the feature will be removed entirely

Since we are to go ahead with version numbers I think I will also change the implementation to automatically hard deprecate by comparing version numbers, if the hard deprecate version number is given. This change will then mean that the hard deprecate argument disappears and that a developer, for a deprecation where the entire schedule is known, will not have to touch the decorator between initial deprecation and deletion.

Thanks for the good suggestions.

Plan now is to split the `only_one_of` decorator out into a separate branch, implement the changes and remove the draft tag from this.

PR #35, opened 2022-03-03
Dilute gas calibration

AnnaWiniwarter:
To enable calibration using very dilute gas mixtures assuming the total gas flux through the capillary is the same as for the pure carrier gas, and proportional to the concentration of the analyte. 
I modified MSInlet.gas_flux_calibration() to enable this. I am aware this is not a very pretty way since you intended to create a class for gas mixtures and then calculate their flux through the capillary in a different way, but this seemed to be the most straightforward modification for me. Happy to change it if you have better ideas.
(Since I don't have data for dilute gas mixtures available I tested it with air and the results seem plausible.)

### DIFF
--- a/src/ixdat/techniques/ms.py
+++ b/src/ixdat/techniques/ms.py
@@ -415,6 +415,8 @@ class MSInlet:
         tspan=None,
         tspan_bg=None,
         ax=None,
+        carrier_mol=None,
+        mol_conc=None,
     ):
         """
         Args:
@@ -425,6 +427,12 @@ class MSInlet:
             tspan_bg (iter): Optional timespan at which the signal is at its background.
             ax (matplotlib axis): the axis on which to indicate what signal is used
                 with a thicker line. Defaults to none
+            carrier_mol (str): The name of the molecule of the carrier gas if
+                a dilute analyte is used. Calibration assumes total flux of the
+                capillary is the same as the flux of pure carrier gas. Defaults
+                to None.
+            mol_conc (float): concentration of the dilute analyte in the carrier gas
+                in ppm. Defaults to None. 
 
         Returns MSCalResult: a calibration result containing the sensitivity factor for
             mol at mass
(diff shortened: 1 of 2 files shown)


src/ixdat/techniques/ms.py
```diff
@@ -425,21 +427,44 @@ def gas_flux_calibration(
             tspan_bg (iter): Optional timespan at which the signal is at its background.
             ax (matplotlib axis): the axis on which to indicate what signal is used
                 with a thicker line. Defaults to none
+            carrier_mol (str): The name of the molecule of the carrier gas if
+                a dilute analyte is used. Calibration assumes total flux of the
+                capillary is the same as the flux of pure carrier gas. Defaults
+                to None.
+            mol_conc (float): concentration of the dilute analyte in the carrier gas
```
ScottSoren, line 434: I would greatly prefer `mol_conc` to be a simple mol fraction (i.e., 1 ppm corresponding to `mol_conc=1e-6`). It might be hard to remember it's in ppm when you have, say, a 1% gas mixture (which I'd think of as `mol_conc=0.01`).
Can you either (i) change `mol_conc` to a simple mol fraction, or (ii) rename it to `mol_conc_ppm` ?

ScottSoren (changes requested):
Hi Anna, thanks for this PR! Looks like a great addition! 
I've just made one comment, about the not-necessarily-intuitive unit of `mol_conc`. Once you address that I'll merge it and build it into a new ixdat for pip distribution :)

src/ixdat/techniques/ms.py
AnnaWiniwarter: Hi Soren,
Thanks for the feedback, good point. I decided to go for `mol_conc_ppm`. I went for the unit ppm on purpose, to discourage using this function if the analyte concentration is in the percent range. Having that in the name will only help with this. :)

ScottSoren: Good point. I like it with the rename.

PR #40, opened 2022-03-14
Def calc l cap effective

AlexanderKrabbe:
I have added a function to calculate the effective length of a capillary given a known n_dot per pure gas.

The value is returned and put in a dict {gas:l_cap_gas_eff} such that the mean of the values can be used to calculated an averaged effective length. 

A function to update self.l_cap (float) given l_cap_eff (dict) is not empty. 

Maybe this should be directly in calc_l_cap_eff(), but I didnt want to automatically update the value self.l_cap.

### DIFF
--- a/src/ixdat/techniques/ms.py
+++ b/src/ixdat/techniques/ms.py
@@ -342,14 +342,10 @@ class MSInlet:
         self.T = T
         self.gas = gas  # TODO: Gas mixture class. This must be a pure gas now.
 
-
     def calc_l_cap_eff(
-        self, gas=None, n_dot=None, w_cap=None, h_cap=None,  T=None, p=None
-        ):
+        self, n_dot_measured, gas=None, w_cap=None, h_cap=None,  T=None, p=None):
         """Calculate gas specific effective length of the capillary in [m]
             and add {gas:value} to l_cap_eff (dict)
-
-        Uses Equation 4.10 of Daniel's Thesis.
 
         Args:
             w_cap (float): capillary width [m], defaults to self.w_cap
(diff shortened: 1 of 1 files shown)


src/ixdat/techniques/ms.py
```diff
@@ -335,15 +335,97 @@ def __init__(
...
+        if w_cap is None:
+            w_cap = self.w_cap  # capillary width in [m]
+        if h_cap is None:
+            h_cap = self.h_cap  # capillary height in [m]
+        if T is None:
+            T = self.T
+        if p is None:
+            p = self.p
+        if n_dot is None:
+            n_dot = calc_n_dot_0(gas=gas)
```
ScottSoren, line 374: What's the point of this? If n_dot is not given, then it looks like `l_cap` necessarily becomes just `self.l_cap`.
Maybe better to make `n_dot` a required argument rather than a kwarg. It doesn't make sense to call this function without it, right?

src/ixdat/techniques/ms.py
```diff
@@ -335,15 +335,97 @@ def __init__(
         """
         self.verbose = verbose
         self.l_cap = l_cap
+        self.l_cap_eff = {}
```
ScottSoren, line 338: cool idea! I think some people at a certain mass spec inlet company would be excited to learn whether `l_cap_eff` is different for different molecules.

ScottSoren (changes requested):
Hey man, looks good, but I'm requesting a few changes. The main thing is not to repeat code from `calc_n_dot_0`. Should be easy to fix and get this merged.

src/ixdat/techniques/ms.py
ScottSoren: (Note that in my own code, I way overdo flexibility of function inputs - I forget to remember from the Zen that "there should only be one right way to do things".)

src/ixdat/techniques/ms.py
```diff
@@ -370,6 +452,7 @@ def calc_n_dot_0(
         if p is None:
             p = self.p
 
+
```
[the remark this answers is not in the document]
AlexanderKrabbe: removed

ScottSoren (approved):
Looks good! I'll merge it now.
Note that we're almost done with user_ready. 
I'll merge this into master with another couple things in the next few days, and then we will be on ixdat v0.2.0. A few things may change.

PR #42, opened 2022-03-15
Debug after big merge

ScottSoren:
Hi Kenneth, here is a small follow-up to our big merge (#30) resulting from going through the [tutorials and article repositories](https://github.com/ixdat/ixdat/blob/debug_after_big_merge/docs/source/tutorials.rst) and the reader_testers. This also resulted in a number of changes to tutorials and article repositories, which was done in each case on an **ixdat_v0p2** branch. 

Things of note in this PR

- convention established of using the string `"mol@mass"` for calibration names, or aliases for molecules of a specific isotope. Previously "_" had been used instead of "@" but it turns out this convuses `MSCalibration` which splits on arguments like `"n_dot_H2"` on `"_"` to find the molecule name. 
- A Spectrum contains a 1-D field defined only on the scanning variable, with a tstamp separate in the object. (Previously it was a 2-D shape=(N, 1) field defined on the scanning variable and a one-point time variable)
- `Measurement.tspan` is now a managed attribute, so that the measurement can clear the cache in its setter to ensure that data series come out of the measurement with the measurement's new t=0 point.

After this, there will be another small PR with recent contributions from Anna and Krabbe. They need to be merged in from the user_ready branch.

And then we're ready to "launch" ixdat v0.2.0 with everything based on this master version :)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -122,7 +122,7 @@ for academic publications:
 Join us
 -------
 
-`ixdat`` is free and open source software and we welcome input and new collaborators. Please help us improve!
+``ixdat`` is free and open source software and we welcome input and new collaborators. Please help us improve!
 
-Contact us (sbscott@ic.ac.uk) or just
+Contact us (https://github.com/ixdat/ixdat/discussions or sbscott@ic.ac.uk) or just
 `get started developing <https://ixdat.readthedocs.io/en/latest/developing.html>`_.
--- a/README.rst
+++ b/README.rst
@@ -122,7 +122,7 @@ for academic publications:
 Join us
 -------
 
-``ixdat`` is free and open source software and we welcome input and new collaborators. Please help us improve!
+``ixdat`` is free and open source software and we welcome input and new collaborators. Please help us improve ``ixdat``!
 
 Contact us (https://github.com/ixdat/ixdat/discussions or sbscott@ic.ac.uk) or just
 `get started developing <https://ixdat.readthedocs.io/en/latest/developing.html>`_.
--- a/src/ixdat/measurements.py
+++ b/src/ixdat/measurements.py
@@ -373,6 +373,7 @@ class Measurement(Saveable):
 
     @property
     def tstamp(self):
+        """Float: The unix epoch time used by the measurement as t=0"""
         return self._tstamp
 
     @tstamp.setter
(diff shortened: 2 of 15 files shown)


KennethNielsen:
May I ask what the reasoning was behind this change:

> A Spectrum contains a 1-D field defined only on the scanning variable, with a tstamp separate in the object. (Previously it was a 2-D shape=(N, 1) field defined on the scanning variable and a one-point time variable)

I can sort of see both working when they are to be combined, so I'm just curious.

KennethNielsen:
| And then we're ready to "launch" ixdat v0.2.0 with everything based on this master version :)

KennethNielsen:
> After this, there will be another small PR with recent contributions from Anna and Krabbe. They need to be merged in from the user_ready branch.

Were they merged after you merged the big merge?

README.rst
```diff
@@ -122,7 +122,7 @@ for academic publications:
 Join us
 -------
 
-`ixdat`` is free and open source software and we welcome input and new collaborators. Please help us improve!
+``ixdat`` is free and open source software and we welcome input and new collaborators. Please help us improve!
```
KennethNielsen, line 125: "Please help us improve ixdat!" maybe

src/ixdat/measurements.py
```diff
@@ -363,14 +363,25 @@ def from_component_measurements(
...
                 series_list.append(vseries)
 
-        # Finally, add this series to the dictionary representation and return the object
+        # Finally, add the series to the dictionary representation and return the object
         obj_as_dict["series_list"] = series_list
         return cls.from_dict(obj_as_dict)
 
+    @property
+    def tstamp(self):
+        return self._tstamp
```
KennethNielsen, line 376: Should the property have a docstring?

KennethNielsen (approved):
Looks good to me. Only a few nit picks. Could be merged as is.

README.rst
ScottSoren: yes, haha. 
May not be the right place to ask for more general help ;)

src/ixdat/measurements.py
ScottSoren: added.

ScottSoren:
1. Behind the change was the fact that it felt much more natural to write the spectroelectrochemistry code with a 1-D spectrum
2. Yes. I'll make master the default branch now.
3. Yes. Open source never sleeps. (or something)

I've addressed your two comments and I'll merge this now :)

README.rst
KennethNielsen: Nah, well, it was more like I think it sounded a bit like we have no idea what we are doing, which I think is not quite right ;)

PR #43, opened 2022-03-15
Deprecation decorators

KennethNielsen:
After the discussion in #34 I now consider the deprecation decorator functionality ready for review. As an outcome of the discussion the interface has changed to use specific version, instead of time intervals, to mark hard deprecation and deletion. The interface now is:
```python
def deprecate(
    last_supported_release,
    update_message,
    hard_deprecation_release=None,
    remove_release=None,
    kwarg_name=None,
):
```

which means that when applying the decorator to a class, method or function, the functionality is immediately considered soft deprecated, which means that it will issue warnings. If desired, it is possible to additionally provide `hard_deprecation_release` and `remove_release` to forewarn of when the functionality will become hard deprecated (raising exceptions) and finally be deleted. The two options compose in the warning and exception message, so while it is a little odd, it is possible to set a remove release and no hard deprecation release. But importantly, as requested, it is possible to leave both out to have the feature be soft deprecated indefinitely.

One note on this though. During the [discussion](https://github.com/ixdat/ixdat/pull/34#issuecomment-1065324297) @ScottSoren mentioned this option of indefinitely soft deprecating something. I would actually suggest to reverse this thinking, at least in the way we suggest to deprecate. I would suggest a relative short soft deprecation and a potentially long hard deprecation. The thinking behind this is that both the warning and the exception will contain instructions on how to fix the offending code, but only the exception will force you to.

Examples of use:
```python
@deprecate("0.1.6", "please use `newfunc` instead")
def func(a, b, c=None, d=None, e=None, f=None):
    return "called"

# OUTOUT
#/home/kenneth/venv/ixdat/ixdat/test.py:11: DeprecationWarning: The function 'func' is deprecated, its last supported version being 0.1.6:
#* It will continue to be soft deprecated, issuing warnings, for the foreseeable future
#
#See instructions below on how to update your code to avoid this message:
#please use `newfunc` instead

@deprecate("0.1.6", "please use `newfunc` instead", hard_deprecation_release="0.2.1")
def func(a, b, c=None, d=None, e=None, f=None):
    return "called"


func(1, 2, f=10)

# OUTPUT
#/home/kenneth/venv/ixdat/ixdat/test.py:12: DeprecationWarning: The function 'func' is deprecated, its last supported version being 0.1.6:
#* It will become hard deprecated, raising exceptions rather than issuing warnings, from version 0.2.1
#
#See instructions below on how to update your code to avoid this message:
#please use `newfunc` instead
#  func(1, 2, f=10)


@deprecate(
    "0.1.6",
    "please use `newfunc` instead",
    hard_deprecation_release="0.2.1",
    remove_release="1.0.0",
)
def func(a, b, c=None, d=None, e=None, f=None):
    return "called"


func(1, 2, f=10)

# OUTOUT
#/home/kenneth/venv/ixdat/ixdat/test.py:17: DeprecationWarning: The function 'func' is deprecated, its last supported version being 0.1.6:
#* It will become hard deprecated, raising exceptions rather than issuing warnings, from version 0.2.1
#* It is planned for complete removal in version 1.0.0
#
#See instructions below on how to update your code to avoid this message:
#please use `newfunc` instead
#  func(1, 2, f=10)


@deprecate("0.1.6", "please use `newfunc` instead", kwarg_name="f")
def func(a, b, c=None, d=None, e=None, f=None):
    return "called"


func(1, 2, f=10)

# OUTOUT
#/home/kenneth/venv/ixdat/ixdat/test.py:12: DeprecationWarning: The argument named 'f' in function 'func' is deprecated, its last supported version being 0.1.6:
#* It will continue to be soft deprecated, issuing warnings, for the foreseeable future
#
#See instructions below on how to update your code to avoid this message:
#please use `newfunc` instead
#  func(1, 2, f=10)


@deprecate("0.1.6", "please use `MyNewClass` instead")
class MyClass:
    ...


MyClass()

# OUTPUT
#/home/kenneth/venv/ixdat/ixdat/test.py:12: DeprecationWarning: The class 'MyClass' is deprecated, its last supported version being 0.1.6:
#* It will continue to be soft deprecated, issuing warnings, for the foreseeable future
#
#See instructions below on how to update your code to avoid this message:
#please use `MyNewClass` instead
#  MyClass()
```

The decorator is also completely tested (all permutations of arguments), which might seem a little like overkill for a development function, but it also serves as an example for pytest functionality like parametrization and patching.

Another discussion that cropped up somewhere (with @ScottSoren) is the (maybe anti)-pattern of having a `tools.py` module. I'm certainly open for input like this. One definition that I have seen is that `tools.py` is for user facing random stuff and `utils.py` is for developer facing random stuff. If we were to follow that, this should certainly go in `utils` instead.

Let me know what you think.

### DIFF
--- a/src/ixdat/tools.py
+++ b/src/ixdat/tools.py
@@ -1,6 +1,15 @@
 """This module contains general purpose tools"""
+import inspect
+import warnings
+from functools import wraps
 
 import numpy as np
+from packaging import version
+
+from ixdat.exceptions import DeprecationError
+from ixdat import __version__
+
+warnings.simplefilter("default")
 
 
 def thing_is_close(thing_one, thing_two):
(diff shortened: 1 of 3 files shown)


src/ixdat/tools.py
```diff
@@ -71,3 +80,159 @@ def list_is_close(list_one, list_two):
...
+    """Return a deprecation message
+
+    Args:
+        identity (str): The identity of the callable being deprecated e.g.
+            "function 'myfunction'"
+
+    All other arguments are as in :func:`deprecate`
+
+    """
+    property_part = f"argument named '{kwarg_name}' in " if kwarg_name else ""
```
ScottSoren, line 102: Why not just make the `property_part` part of `identity` in the outer scope?

src/ixdat/tools.py
```diff
@@ -71,3 +80,159 @@ def list_is_close(list_one, list_two):
...
+                bound_arguments = callable_signature.bind(*args, **kwargs).arguments
+
+            # If something deprecated is being used, issue the warning or exception
+            if kwarg_name is None or kwarg_name in bound_arguments:
+                if hard_deprecation_release is not None and version.parse(
+                    __version__
+                ) >= version.parse(hard_deprecation_release):
+                    raise DeprecationError(compound_message)
+                else:
+                    warnings.warn(compound_message, DeprecationWarning, stacklevel=2)
```
ScottSoren, line 229: what is `stacklevel`? A comment here could be useful.

src/ixdat/tools.py
```diff
@@ -71,3 +80,159 @@ def list_is_close(list_one, list_two):
...
+
+        # Get the argument signature of the callable
+        callable_signature = inspect.signature(callable_)
+
+        @wraps(callable_)
+        def inner_function(*args, **kwargs):
+            # Form bound arguments, so both args and kwargs gets mapped to their name
+            bound_arguments = {}
+            if kwarg_name:
+                bound_arguments = callable_signature.bind(*args, **kwargs).arguments
```
ScottSoren, line 220: Do I understand right that this binding `*args` and `**kwargs` to the callable's signature is there to enable deprecation of a positional argument? If we were only deprecating use of `**kwargs`, then it would siffice for the condition below to just be `if kwarg_name in kwargs`, right?
I can't think of a situation where a positional argument would be deprecated without deprecating the function, in which case this might be overkill. But such a situation will probably come and we'll be happy for it :)

tests/unit/test_tools.py
```diff
@@ -0,0 +1,200 @@
+"""Tests for tools.py"""
```
ScottSoren, line 1: Very cool testing! 
I've read through it and can follow it all, though I could not have written it myself without hundreds of hours of study. 
Looks like every aspect of deprecation is fully tested.

ScottSoren (approved):
Looks awesome! And good to merge to **main**.
I've left a few comments.

src/ixdat/tools.py
```diff
@@ -71,3 +80,159 @@ def list_is_close(list_one, list_two):
...
+                "1.4.0",
+                "2.0.0",
+                kwarg_name="kwarga"
+            )
+            def mymethod(cls, arg, kwarga=None, kwargb=None):
+                ...
+
+    .. note::
+       In the example above, when this decorator is applied to a class method or
+       static method, it must be applied as the first decorator (closest to the def).
```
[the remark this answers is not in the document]
KennethNielsen: I don't know. I kind of thought that I was already over engineering when I extended it to static and class methods. I will check.

src/ixdat/tools.py
KennethNielsen: Ah, agreed! The stack level refers to the call stack. This is a way to make sure that the reference in the warning text about where the warning was issued, will refer to the correct place in the decorated function and not to the place here in the inner function, which would be useless. I will add a comment.

src/ixdat/tools.py
KennethNielsen: Ah, there actually is a good reason for this. It's one of those cases where Python dynamic nature is ... challanging. I've actually tried and failed to solve this problem before. The problem is that you may reach a kwarg as an arg i.e if we want to deprecate `c` in the following function:
```python
def myfunc(a, b, c=None):
    pass

myfunc(1, 2, 3)
```
is allowed and will provide a value for `c` as an arg, no a kwarg. Therefore this trick is needed.

tests/unit/test_tools.py
KennethNielsen: The question is if it is readable enough and the only thing missing is prior knowledge. I did spend some time on it, to make it serve as an example. If not now, let me know later. Maybe we should do a small testing seminar at some point and then you can see afterwards if you can read the tests and if not I will update them. Even with test code there are questions with how far one goes to maximize code re-use at the right level, which has a tendency to make code harder to read. So that may be a problem here as well.

src/ixdat/tools.py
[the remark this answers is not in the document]
KennethNielsen: It does. But as with the other "special" method types, it needs to be applied closest to the def.

This works:
```python
from ixdat.tools import deprecate

class A:

    def __init__(self):
        self._a = 8

    @property
    @deprecate("1.2.3", "Just no")
    def a(self):
        return self._a

    @a.setter
    @deprecate("1.2.3", "Just no")
    def a(self, value):
        self._a = value


a = A()
print(a.a)
a.a = 10
print(a.a)
```

I will update the docs to show that this works.

tests/unit/test_tools.py
KennethNielsen: I'll resolve this for now, but please re-open if you want to discuss about the readability of the test code.

src/ixdat/tools.py
KennethNielsen: Or maybe it is even cleaner to move it the other way. I will figure it out.

KennethNielsen:
@ScottSoren thanks for the review. Good observations in there. I think I should have addressed all of them now.

src/ixdat/tools.py
KennethNielsen: This particular problem is of course prevented by the use of kwarg_only arguments, which you have developed the excellent habit of using to great effect, but I thought it was better if the decorator covered the other use case as well.

src/ixdat/tools.py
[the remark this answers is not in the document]
ScottSoren: Awesome. that will be useful. 
ixdat v0.2+ uses `Measurement.add_calibration(my_cal)` instead of `Measurement.calibration = my_cal` , so I can make a dummy property and deprecate its setter to inform the user :)

tests/unit/test_tools.py
ScottSoren: It **is** readable (in spite of also looking like magic) was my point :)

ScottSoren:
Thanks for the extra comments. Looks good! Merging.

PR #46, opened 2022-03-16
Pin dev dependencies, extra invoke task and black style fixes

KennethNielsen:
I'm planning on some more improvements around tooling. This is step one. The PR contains a couple of small items:

1. I added a new `dependencies` task (alias dep) to upgrade pip, install dependencies and dev dependencies
2. I pinned the version of the dev tools that we use for checks, because it is annoying to fight over differences in this sort or tooling i.e. some code might raise a flake8 warning on one machine and not on another for the same PR due to a version bump on one of the. It is just easier to pin them and upgrade every once in a while.
3. I added a `pyproject.toml` to be able to configure `black` which will now adhere to the BDFL determined one-true line length or 89 chars
4. I updated all files with black fixes as per the new pinned version of black

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,2 @@
+[tool.black]
+line-length = 89
--- a/src/ixdat/techniques/ms.py
+++ b/src/ixdat/techniques/ms.py
@@ -201,9 +201,7 @@ class MSMeasurement(Measurement):
         """
         t, S = self.grab_signal(mass, tspan=tspan, include_endpoints=True)
         if tspan_bg:
-            t_bg, S_bg_0 = self.grab_signal(
-                mass, tspan=tspan_bg, include_endpoints=True
-            )
+            t_bg, S_bg_0 = self.grab_signal(mass, tspan=tspan_bg, include_endpoints=True)
             S_bg = np.mean(S_bg_0) * np.ones(t.shape)
         else:
             S_bg = np.zeros(t.shape)
@@ -470,9 +468,7 @@ class MSInlet:
         self.T = T
         self.gas = gas  # TODO: Gas mixture class. This must be a pure gas now.
 
-    def calc_n_dot_0(
-        self, gas=None, w_cap=None, h_cap=None, l_cap=None, T=None, p=None
-    ):
+    def calc_n_dot_0(self, gas=None, w_cap=None, h_cap=None, l_cap=None, T=None, p=None):
         """Calculate the total molecular flux through the capillary in [s^-1]
 
         Uses Equation 4.10 of Trimarco, 2017. "Real-time detection of sub-monolayer
@@ -512,7 +508,7 @@ class MSInlet:
         p_1 = p
         lambda_ = d  # defining the transitional pressure
         # ...from setting mean free path equal to capillary d
-        p_t = BOLTZMAN_CONSTANT * T / (2 ** 0.5 * pi * s ** 2 * lambda_)
+        p_t = BOLTZMAN_CONSTANT * T / (2**0.5 * pi * s**2 * lambda_)
         p_2 = 0
         p_m = (p_1 + p_t) / 2  # average pressure in the transitional flow region
         v_m = (8 * BOLTZMAN_CONSTANT * T / (pi * m)) ** 0.5
--- a/tasks.py
+++ b/tasks.py
@@ -1,4 +1,29 @@
-"""Definition of invoke tasks"""
+"""Definition of invoke tasks
+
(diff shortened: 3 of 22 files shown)


pyproject.toml
```diff
@@ -0,0 +1,2 @@
+[tool.black]
+line-length = 89
```
ScottSoren, line 2: YES! But this duplicates the same information on line 3 of `setup.cfg` ... why are both needed?

ScottSoren (approved):
Yes! Love it! I'm tired of black installed on one computer splitting up lines and black on another computer folding them back, and it looks like this will solve it. It'll also give an easy way to get someone to fix their python environment (as I recently had to walk a collaborator in China through over email). 
All I could use is a bit more documentation on how to use it - a module docstring on `tasks.py` or and/making sure it's clear enough in `tools.rst`.

ScottSoren:
Oh, it looks like after #43, this has a conflict with the main branch. Could you please do the merge?

pyproject.toml
KennethNielsen: The one in `setup.cfg` only applies to flake8, this is for black. Apparently no-one has been able to agree on the one true place to store such settings or a way to share common settings.

KennethNielsen: However I now realize that I am missing a bullet point in TOOLS.rst about this settings file. I will add it.

tasks.py
```diff
@@ -193,13 +194,33 @@ def clean(context, dryrun=False):
...
+                    print("DELETE FILE:", cleanpath)
+                    if not dryrun:
+                        cleanpath.unlink()
+
+
+@task(aliases=["deps"])
+def dependencies(context, user=False):
+    """Install development and normal dependencies
+
+    See docstring of :func:`flake8` for explanation of `context` argument
```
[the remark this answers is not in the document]
KennethNielsen: Yeah, good point. It is in TOOLS.rst, but it can't hurt to put it in the module docstring.

KennethNielsen:
I implemented the changes, but also added a few more changes as suggested in the review, also in TOOLS.rst, which means that this should probably get a second review BUT I should also resolve the conflicts, which I would rather not do until #47 is merged, since it also touches tasks.py. So when #47 is merged, I will fix this one up and request a second review.

tasks.py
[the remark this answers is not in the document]
ScottSoren: Great. Thanks! I had forgotten the 1:1 between `invoke` commands and `@task`-decorated functions here. Clearer now :)

ScottSoren:
I've merged #47 , and this PR has only gotten better :) . Thanks for the additions to tools.rst!
Good to merge.

KennethNielsen:
I merge upstream/main in to fix the conflicts, but a few more things showed up. I'd like a chance to have a look at the diff myself first before anyone else uses time on it so DO NOT review or MERGE yet.

KennethNielsen:
It's ready for re-review. It should be really small, as almost nothing has changed and these are mostly style fixes. A few things:

- I merged in upstream/main to resolve the conflicts. There were trivial conflicts in tasks.py and non-trivial ones in ms.py, so I choose the easy and safe approach there and used the upstream/main version and black formatted it again. Nonetheless, that is one that I would request a look at
- I found a minor bug in the recently merged addition of constants from @AlexanderKrabbe. Apparently, the molecule diameter had been mistyped as **A - B** and not **Ae-B**, so also please have a look at constants.py.

src/ixdat/techniques/ms.py
```diff
@@ -512,8 +513,7 @@ def update_l_cap(self, gases=[]):
 
         return self.l_cap
 
-    def calc_n_dot_0(
-        self, gas=None, w_cap=None, h_cap=None, l_cap=None, T=None, p=None):
+    def calc_n_dot_0(self, gas=None, w_cap=None, h_cap=None, l_cap=None, T=None, p=None):
```
ScottSoren, line 516: Looks good. It's got both the new method above and this method definition black-formatted.

tasks.py
```diff
@@ -1,4 +1,29 @@
-"""Definition of invoke tasks"""
+"""Definition of invoke tasks
+
+All functions in this file that is decorated with the `@task` decorator,
+constitutes an invoke task, which can be execute from the command line by
+calling invoke with the name of the task, which is the same as the name of the
```
ScottSoren, line 5: awesome.

ScottSoren (approved):
Hi Kenneth, looks good! Comments here are just complementing the PR :) .
I'll merge. Thanks for this.

PR #47, opened 2022-03-17
GitHub actions

KennethNielsen:
This PR aims to setup continuous integration (CI) for ixdat via [Github Actions[(https://docs.github.com/en/actions).

Github Actions is a framework for having certain actions done automatically on certain events. It is heavily used for CI, but it is by no means exclusive to that. It is e.g. possible to setup actions to e.g. make releases triggered by a tag, or to do certain things when PRs are opened, like having a bot comment if the CI fails or similar. But for now I setup just the CI. The CI checks the same items as `invoke checks` does, which now is:

- That the tests pass
- That code and tests passes flake8
- That code, tests and development scripts are black formatted

The checking of black formatted code was added in this PR along with another task for black formatting everything.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,49 @@
+# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: Continuous integration
+
+# Run the CI on any direct push to a branch on this repository or on pull request
+# branches
+on:
+  push:
+    branches:
+      - '**'
+  pull_request:
+    branches:
+      - 'main'
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+        if [ -f requirements-dev.txt ]; then pip install -r requirements-dev.txt; fi
+    - name: Set PYTHONPATH
+      run: |
+        echo "PYTHONPATH=src" >> $GITHUB_ENV
+    - name: Lint with flake8
+      run: |
+        # stop the build if there are Python syntax errors or undefined names
+        invoke lint
+    - name: Test with pytest
+      run: |
+        invoke test
+    - name: Check black formatting of code, tests and development scripts
+      run: |
+        invoke check-code-format
+
--- a/tasks.py
+++ b/tasks.py
@@ -11,6 +11,8 @@ from subprocess import check_call, CalledProcessError, check_output, DEVNULL
 
 THIS_DIR = Path(__file__).parent
 SOURCE_DIR = THIS_DIR / "src" / "ixdat"
+TESTS_DIR = THIS_DIR / "tests"
+DEV_SCRIPTS_DIR = THIS_DIR / "development_scripts"
 # Patterns to match for files of directories that should be deleted in the clean task
 CLEAN_PATTERNS = ("__pycache__", "*.pyc", "*.pyo", ".mypy_cache")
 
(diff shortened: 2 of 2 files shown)


tasks.py
```diff
@@ -11,6 +11,8 @@
 
 THIS_DIR = Path(__file__).parent
 SOURCE_DIR = THIS_DIR / "src" / "ixdat"
+TESTS_DIR = THIS_DIR / "tests"
+DEV_SCRIPTS_DIR = THIS_DIR / "development_scripts"
```
ScottSoren, line 15: I don't think that develompent_scripts should be looked at by CI tools. It is the one version-controlled place that exists purely for the whimsical convenience of the ixdat developer.

Ah, but I see below you only run black on DEV_SCRIPTS_DIR. I think that's okay. Maybe a comment on this line saying so?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,49 @@
...
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+        if [ -f requirements-dev.txt ]; then pip install -r requirements-dev.txt; fi
+    - name: Set PYTHONPATH
+      run: |
+        echo "PYTHONPATH=src" >> $GITHUB_ENV
+    - name: Lint with flake8
```
ScottSoren, line 39: I take it this will only run flake8 on src/ and texts/ as specified in tasks.py? A comment here would help.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,49 @@
...
+      - 'main'
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
```
ScottSoren, line 23: That's a lot of pythons - how long does it take?

ScottSoren (approved):
Looks good! 
I would be worried about it slowing down a push, since I think it's good practice to push often to back up local work, and that there should be no barrier at all to doing so on a feature branch. So long as the push is instant and these CI tools happen on github without making the user wait (for all but the main branch), I'm happy.

KennethNielsen:
> Looks good!
I would be worried about it slowing down a push, since I think it's good practice to push often to back up local work, and that there should be no barrier at all to doing so on a feature branch. So long as the push is instant and these CI tools happen on github without making the user wait (for all but the main branch), I'm happy.

Well this does satisfy your opinion on this, but I think I would object slightly to the idea, since that would definitely get in the way of something like git hooks. The way I view it, HD fails happens, but are infrequent, so the major thing that version control systems should guard against, if we are talking about backup, is user induced losses i.e. Error 40. For that reason, I think **commit** (which guards against that) should be fast, but **push** (which guards against HD failure) should be allowed to take seconds. This will likely mean that commit is suitable for all the time and push some times per day, possible at natural breaks, lunch, coffee run etc. etc. which I think is a reasonable compromise between what pre-commits adds of value and the risk of losing what you worked on the last few hours due to HD failure once every few years.

tasks.py
KennethNielsen: Good point and I will add it.

.github/workflows/ci.yml
KennethNielsen: Yeah, I was debating with myself about how much to comment this file WRT to what it runs. Since the `run` part specifically refers `invoke` I thought this would be sufficient, but I will add the comment no problem.

.github/workflows/ci.yml
KennethNielsen: Well those are the Pythons we support. Usually, you will run CI on all the Python version you support, to make sure that you didn't inadvertently used syntax which is exclusive to newer Pythons or that one of our dependencies have stopped supporting an older version. To answer your question, minutes, usually, a few  and rarely no more than you can count on fingers. It is a little difficult to answer precisely, because using the free service, we are of course on shared HW which by design has a high load, so sometimes you have to "wait your turn". However, I have never observed this practically to be a problem, because the cases where you really want CI results visible on GH, is when you are handing a PR over for review, which almost always involves another person, which again usually have a hand off latency of more than a few minutes ;)

The hidden agenda here is, that with the ability to run CI (especially tests) on all supported Pythons automatically on GH's systems, we essentially remove the need for tox, which mostly served that purpose, just locally, where actually getting all those Pythons installed is a hassle anyway. So if this is merged, I will shortly propose to remove tox, to simplify tooling somewhat.

KennethNielsen:
@ScottSoren I consider this ready to merge, but will leave it open for you to merge, to maintain proper workflow and allow you to see answers to certain questions.

ScottSoren:
Yes. Looks good!
I'll just answer that the philosophy of frequent seamless pushing is not so much out of concern for HD failure, but more HD unavailability. If I work at home today and pushing isn't an instinctive twitch, there's a good chance that tomorrow I'll be at work and not be able to do what I wanted since I left my laptop at home and my git repositories are (unlike everything else) not in dropbox. So I'm glad that push will still be seamless for the user :)

PR #48, opened 2022-03-17
tidy_up_doc_strings

AlexanderKrabbe:


ScottSoren:
Awesome, thanks! I will merge this into new_contributions, and then into main!

Issue #20, opened 2021-11-24
Problem in .as_cv when using an ECMS object that is created by adding ECMS object + EC object


spoeka22:
I tried to find a workaround to using the noisy EC data in Zilien dataset by doing something like the following:
zilien_obj = Measurement.read( "example_data", reader="zilien")
bio_obj =  Measurement.read( "example_ec data", reader="biologic")
combi_obj = zilien_obj + bio_obj

This results in an ECMS object, where I can select the less noisy "<I>/mA" as J_str to get a plot vs time:combi_obj.plot_measurement(J_str="<I>/mA" ) 

However, when trying to first convert to cv with combi_obj.as_cv() and then using combi_obj.plot_vs_potential() something breaks.
(1) initially the cycles are NoneType objects
(1) When using redefine_cycle, the cycles are somehow broken into very small fragments (in the case I tested 3 cycles became 10)
(2) combi_obj.plot_vs_potential(J_str="<I>/mA") gives following error: AttributeError: 'Line2D' object has no property 'j_str'

spoeka22:
this might be related to issue #14

ScottSoren:
I think #14  is a much worse headache... it's been solved in a big restructuring on the [master] branch. Merging that into [user_ready] will be tough.

spoeka22:
See comment to #17. Problem solved by not introducing the conflict of having to deal with concurrent EC data from different sources (biologic file and zilien file)

spoeka22:
I'm reopening this issue, because I realised the problem also occurs when just combining several biologic files (a CVA and some CP measurements), so it's not only because of incompatibility with EC data from different importers.

ScottSoren:
This is addressed in #30 . I'm assigning @KennethNielsen because as far as I can tell #30 is still on his desk :) 

As input to that, @spoeka22 , if you have time on your hands, it would be great to hear if the problem raised in this issue is solved if you [run ixdat from](https://ixdat.readthedocs.io/en/latest/developing.html) the [backend_fixes branch](https://github.com/ixdat/ixdat/tree/backend_fixes).

AnnaWiniwarter:
Hej Soren, no unfortunately it still doesnt work. It breaks completely differently now, though:
If I import the zilien file as `MSMeasurement` and then try to add the `ECMeasurement` objects, I get following error:
```
  File "\ixdat\src\ixdat\techniques\ec_ms.py", line 70, in as_dict
    if self.calibration:

[...]

Issue #29, opened 2021-12-02
EC calibration seems to get lost when loading MS calibration


spoeka22:
Maybe I'm just calling functions in the wrong way/wrong order, but if I add an MSCalResults object to an ECMSMeasurement object using the following, the previously added EC calibration (for RHE potential and electrode area) get lost.
meas is an ECMSMeasurement:
meas.calibrate(RE_vs_RHE=0, A_el=0.196, R_Ohm=0)
cal_co2 = MSCalResult(name="CO2_M44", mol="CO2", mass="M44", F=F_blank,
            cal_type="internal", )
meas.calibration = ECMSCalibration(ms_cal_results=[calibration])
-> now the EC calibration is gone.

if I add it again, when adding the ms calibration, it works:
meas.calibration = ECMSCalibration(ms_cal_results=[calibration],RE_vs_RHE=0, A_el=0.196)

ScottSoren:
Hey Anna,
Yeah, this is stupid. It's because RE_vs_RHE etc live in meas.calibration. You already found the workaround :D

A real fix is on the way: from the master branch Soon measurements will actually have a list of calibrations, and meas.calibration will group the data in all calibrations in the list. (You'll have to use add_calibration(...) rather than calibration=...)

ScottSoren:
@AnnaWiniwarter , is this okay to close after #50 ?

Issue #22, opened 2021-11-30
ECMScalibration and ECMSCyclicVoltammogram seem to not be compatible


spoeka22:
I tried to run a calibration over a selected t_span of an ECMSCyclicVoltammogram object, lets callit ecms_cvmeas:
ecms_cvmeas.ecms_calibration('CO2', 'M44', n_el=+2, tspan=[700, 800], tspan_bg=[600,700])
results in following warning: 
ixdat\src\ixdat\techniques\ec_ms.py:121: RuntimeWarning: invalid value encountered in double_scalars
  F = Y / n
and an output without F: MSCalResult(name=CO2_M44, mol=CO2, mass=M44, F=nan)

if on the other hand I run the calibration before converting the ECMSMeasurement to an ECMSCyclicVoltammogram object, the calibration works fine, if I then use cvmeas=ecms_cvmeas.as_cv() the (calibrated ?) CMSCyclicVoltammogram object (cvmeas) doesnt have any cycles (i.e. if I then call cvmeas[0] (or higher indices), it returns None), but the data itself is there.

I can work around this by just selecting the cv cycles by time instead of cycle, of course, but it's just less convenient. :)

spoeka22:
Also, slightly relayed, I noticed that in the code both voltammagram (with an "a", incorrect spelling) and voltammogram (with an "o", the correct spelling) are being used. I understand that changing all the incorrectly spelled occurrences might be a problem for backwards compatibility, but having both spellings in there in different places is a bit confusing. :)

KennethNielsen:
Thanks for reporting this. Can you share a script (and the needed data files) that reproduces the issue?

KennethNielsen:
Can I persuade you to report the spelling as a separate issue?

> ... I understand that changing all the incorrectly spelled occurrences might be a problem for backwards compatibility, but having both spellings in there in different places is a bit confusing. :)

There are usually ways around that.

spoeka22:
@KennethNielsen I uploaded some code plus data here: https://github.com/spoeka22/info_issue_reports
Hope that works for you.
Note that I've been using the [fuzhou] branch when I first observed this issue, but I also tested on [user_ready] now and it behaves the same.

ScottSoren:
@AnnaWiniwarter , is this okay to close after #50 ?

AnnaWiniwarter:
Yes, this is solved for me after #50 (with new documentation on the importance of using `.redefine_cycle()`)

PR #50, opened 2022-03-20
Lets deprecate

ScottSoren:
I expect this is my final PR on the way to ixdat v0.2.0 release. 

I'm requesting eyes from

-  @KennethNielsen for the use of deprecation decorators and changelog workflow, and 
-  @AnnaWiniwarter to see if the API changes are adequately described for use in e.g. EC-MS calibration, since I'm claiming it resolves the Issues related to that (including #20, where what had last time given an error will now give a `DeprecationWarning`).

The PR does the following:

- Introduces a changelog system. The instructions are described in docs/developing and at the top of NEXT_CHANGES.rst
- Describes the changes to the API going from ixdat v0.1.x to ixdat v0.2.0 in NEXT_CHANGES.rst. I will move this, after your review, to CHANGES.rst for release of v0.2.0. 
- Implements deprecation (introduced in #43). These are all set to give a warning in ixdat v0.2.x and raise an error in ixdat v0.3. The decorators are extensive so that people can for the most part keep running their scripts built in ixdat v0.1.x with . I think I've mentioned all the deprecations in NEXT_CHANGES.rst
- Fixes a bug by which it was sometimes impossible to look up a cache'd `series` in a measurement using its `series.name` as the key. 

In addition to the code here, I've updated all the other repositories that use ixdat so that they are fully ready for ixdat v0.2.0 release. In each of the following, I've made an **ixdat_v0p1** branch and merged the **ixdat_v0p2** branch into **main**. The scripts on the **ixdat_v0p1** branches are able to run with both ixdat v0.1.9 and v0.2.0 (code in this PR), with a lot of deprecation errors issued in the later. The README on the **main** branches describe the requirement of ixdat v0.2+ and have links to the **ixdat_v0p1**. This applies for:

- The tutorials repository: https://github.com/ixdat/tutorials
- The EC-MS article repository at https://github.com/ScottSoren/Huang2021
- The EC-MS article repository at https://github.com/ScottSoren/pyCOox_public

Looking forward to your feedback and to saying goodbye to ixdat v 0.1.x ! :)


KennethNielsen:
Looks like a lot of great work and important work. I will review shortly.

The tests fails due to a odd dependency problem I encountered during #46. The development dependencies list `setuputils`, but that package is a Python 2 only dependency, which doesn't make any sense for ixdat. My best guess is that at the `requirements-dev.txt` was created in retrospect in an environment that was already setup and working and therefore this just registered as a weird PyPI error when upgrading. In any case, I changed to dependency to `setuputils3` (which is the Python 3 versin of the former), which is my best guess at what it was supposed to be. So if you merge #46 and merge upstream/main into this, it should pass the tests (except for any style issues possibly introduced in this PR).

ScottSoren:
After merging in `main` and making a few tiny fixes, it passes the github CI tools :)

NEXT_CHANGES.rst
```diff
@@ -0,0 +1,102 @@
...
+  - ``v_name`` (replaces ``V_str``)
+  - ``j_name`` (replaces ``J_str``)
+  - ``E_name`` (replaces ``E_str``)
+  - ``I_name`` (replaces ``I_str``)
+
+  The old property names are deprecated.
+
+- Renamed keyward in ``MSMeasurement.plot_flux()``:
+
+  - ``remove_backround`` (replaces ``removebackground``
```
AnnaWiniwarter, line 42: it should be ``remove_background``, right? Is the typo also in the code?

src/ixdat/measurements.py
```diff
@@ -439,6 +440,26 @@ def c_ids(self):
     def add_calibration(self, calibration):
         self._calibration_list = [calibration] + self._calibration_list
 
+    @property
+    @deprecate(
+        "0.1",
+        "At present, ixdat measurements have a `calibration_list` but no compound "
+        "`calibration`, and the property just returns the first from the list.",
```
KennethNielsen, line 447: maybe ".. and **this** property"

src/ixdat/measurements.py
```diff
@@ -439,6 +440,26 @@ def c_ids(self):
     def add_calibration(self, calibration):
         self._calibration_list = [calibration] + self._calibration_list
 
+    @property
+    @deprecate(
```
KennethNielsen, line 444: Alright so, I was sort of expecting a problem around writing the keyword arguments out in full or not when I wrote it. I think, that since the is breaching multiple lines any, that I would as a courtesy to other developers, because the deprecate calls are not exactly self explanatory. Long story short, I think that for the ones that breach multiple lines anyway, I would use the keywords to make it easier to read.

This also helps a little with making sure that the message, which may be multi-line, is easier to separate from the next arguments. Consider how small the difference is between these two:
```python
@deprecate(
    "0.1",
    "At present, ixdat measurements have a `calibration_list` but no compound "
    "`calibration`, and the property just returns the first from the list.",
    None,
)

@deprecate(
    "0.1",
    "At present, ixdat measurements have a `calibration_list` but no compound ",
    "`calibration`, and the property just returns the first from the list.",
    None,
)
```
where the latter is a bug that fills in a nonsensical `hard_deprecation_release`. Whereas if we do:
```python
@deprecate(
    last_supported_version="0.1",
    update_message=(
        "At present, ixdat measurements have a `calibration_list` but no compound "
        "`calibration`, and the property just returns the first from the list."
    )
    hard_deprecation_release=None,
)
```
that risk is seriously mitigated.

I will not comment on this more than once.

src/ixdat/measurements.py
```diff
@@ -589,9 +610,25 @@ def __getitem__(self, key):
...
         # Finally, wherever we found the series, cache it and return it.
         # step 3.
-        self._cached_series[key] = series
+        self._cache_series(key, series)
         return series
 
+    def _cache_series(self, key, series):
+        """Cache `series` such that it can be looked up with its name or with `key`."""
+        self._cached_series[key] = series  # now it can be looked up with by `key`
+        # If the name of the series is not `key`, we can get in a situation where
```
KennethNielsen, line 619: I'm a little in doubt about this change. Is this change changing what aliases may or may not be used for? I was under the assumption that aliases were either defined by readers or calibrations and were hence forth static, but this certainly changes that. Is that on purpose? And/or am I understanding the concept of alias wrong? Additionally, how would this compare to simply adding the series to `_cached_series` under their name also? Or will that somehow mess up because because of a possibility for more than one with the same name?

src/ixdat/plotters/ec_plotter.py
```diff
@@ -12,18 +13,24 @@ def __init__(self, measurement=None):
         """Initiate the ECPlotter with its default Meausurement to plot"""
         self.measurement = measurement
 
+    @deprecate("0.1", "Use `v_name` instead.", "0.3", kwarg_name="V_str")
```
KennethNielsen, line 16: Ha. I had never, annoyingly, consider how these would stack. Presumably the `@wraps` decorator copies the necessary callable signature over, otherwise it wouldn't work. Have you tested that it works both for more than the first argument name "j_color", because I don't test that.

src/ixdat/plotters/ec_plotter.py
```diff
@@ -12,18 +13,24 @@ def __init__(self, measurement=None):
...
+    @deprecate("0.1", "Use `j_name` instead.", "0.3", kwarg_name="J_str")
```
KennethNielsen, line 17: Here BTW we might excuse ourselves from writing the kwarg names out to keep it on one line.

src/ixdat/plotters/ec_plotter.py
```diff
@@ -37,18 +44,20 @@ def plot_measurement(
...
             j_name (string): The name of the ValueSeries to plot on the right y-axis.
                 Defaults to measurement.J_str, which for an ECMeasurement is the name
                 of its most normalized/correct current.
-            axes (list of matplotlib.Axis): Two axes to plot on, if not the default
-                new twinx()'d axes. axes[0] is for V_str and axes[1] for J_str.
-            V_str (str): DEPRECIATED. now v_name
-            J_str (str): DEPRECIATED. now j_name
             v_color (str): The color to plot v_name. Defaults to black.
             j_color (str): The color to plot j_name. Defaults to red.
+            V_str (str): DEPRECATED
```
KennethNielsen, line 57: Maybe, you should keep the new names here still, so that it is deducible from the API docs only, without running the code.

src/ixdat/plotters/ecms_plotter.py
```diff
@@ -12,6 +15,13 @@ def __init__(self, measurement=None):
         self.ec_plotter = ECPlotter(measurement=measurement)
         self.ms_plotter = MSPlotter(measurement=measurement)
 
+    @deprecate("0.1", "Use `v_name` instead.", "0.3", kwarg_name="V_str")
```
KennethNielsen, line 18: :eyes: "Let's deprecate" :notes:

src/ixdat/plotters/ecms_plotter.py
```diff
@@ -64,13 +81,18 @@ def plot_measurement(
...
                 background signals if available. Defaults to (not logplot)
+            removebackground (bool): DEPRECATED
             unit (str): the unit for the MS data. Defaults to "A" for Ampere
             v_name (str): The name of the value to plot on the lower left y-axis.
                 Defaults to the name of the series `measurement.potential`
             j_name (str): The name of the value to plot on the lower right y-axis.
                 Defaults to the name of the series `measurement.current`
             v_color (str): The color to plot the variable given by 'V_str'
             j_color (str): The color to plot the variable given by 'J_str'
+            V_str (str): DEPRECATED
```
KennethNielsen, line 92: Maybe also here, list the new names

src/ixdat/techniques/cv.py
```diff
@@ -324,6 +325,15 @@ def diff_with(self, other, v_list=None, cls=None, v_scan_res=0.001, res_points=1
         return diff
 
 
+class CyclicVoltammagram(CyclicVoltammogram):
+
+    # FIXME: decorating the class itself doesn't work because the callable returned
```
KennethNielsen, line 330: Wow. Had not considered that. Nice catch.

KennethNielsen: The reason I was angling to avoid this, was that I was unsure if the actual class name was being used anywhere in (de)/serialization code and so would cause a data hygiene problem. If that is the case, we might just move ahead and hard deprecate immediately.

src/ixdat/techniques/ms.py
```diff
@@ -145,7 +150,11 @@ def grab_flux(
                 If not given, no background is subtracted.
             remove_background (bool): Whether to remove a pre-set background if available
                 Defaults to True.
+            removebackground (bool): DEPRECATED
```
KennethNielsen, line 153: and here, new name. I will not comment on it further.

src/ixdat/techniques/cv.py
KennethNielsen: I will raise an issue on it, since I've documented being able to use it for the class.

KennethNielsen (changes requested):
Looks great. Nice to see that decorator being put to good use.

There are a few typos and some questions in the review. Mostly simple stuff I think. Should be a quick turn around. Just let me know whether any of it warrants a second look.

AnnaWiniwarter (changes requested):
Of the edited files, I only had a look at NEXT_CHANGES.rst (see comments there, Kenneth had more than I did).
There is no mention of the change to calling calibrated molecules now with `mol_list=["CO2@M44"]` instead of old `mol_list=["CO2_M44"]` (and this just results in an error instead of deprecation comment)
Also,   `co_strip.add_calibration(RE_vs_RHE=0, A_el=0.196, R_Ohm=0)` doesn't work, but I cannot really figure out what the new syntax should be if I dont use `.calibration()` anymore. so I couldn't test if issue #29 is solved

Then I tried running some of my old scripts where I encountered errors in the calibration.
First of all, somehow the importing of .csv files exported with some 0.1.x version seems to be broken: 
image
I did not investigate this further, but I think I already ran into this problem when using the new "main" branch after The Big Merge.

If I then import the data just straight from the Zilien file instead and try and do an ECMS calibration, it still breaks when I try to use `.as_cv()` to convert it to an ECMS_Voltammogram object. Instead of choosing one cycle if I call `my_cv[1]` it chooses some part of the entire dataset, and and a different part `my_cv[2]` but both kind of stretching over the entire time of the measurement. 
on the plus side, the result is the same whether I call `.as_cv()` before or after the calibration. That means that issue #22 appears to be solved (it is now possible to use ecms_calibration on a ECMSCyclicVoltammogram object)

finally, I tried to convert a combination of files imported from Zilien and biologic data to a CyclicVoltammogram object (issue #20). this works, but when plotting it somehow plots the current double with a weird time shift: 
Capture2
(top image is if I import full dataset from zilien file, bottom if I combine the MS data from zilien (imported with `MSMeasurement.read()` with EC data from biologic. Note this happens regardless of whether I convert it to a ECMSCyclicVoltammogram object or not.

Hope my comments are somewhat clear.

NEXT_CHANGES.rst
```diff
@@ -0,0 +1,102 @@
+Instructions
+============
+
+Use this as a staging ground for CHANGES.rst. In other words, describe the
+changes and additions to ixdat's API associated with your contribution. The idea is
+that what you write here informs other developers what is on its way and then will be
+copied to CHANGES.rst when the next version of ixdat is distributed. Please include
+links to relevant Issues, Discussions, and PR's on github.
```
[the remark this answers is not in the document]
ScottSoren: Done.

NEXT_CHANGES.rst
```diff
@@ -0,0 +1,102 @@
...
+^^^^^^^^^^^^^^^^
+
+- Renamed measurement technique class: ``CyclicVoltammogram`` (replaces ``CyclicVoltammagram``)
+  The old name is deprecated.
+
+- Renamed properties in ``ECMeasurement`` and inheriting classes:
+
+  - ``v_name`` (replaces ``V_str``)
+  - ``j_name`` (replaces ``J_str``)
+  - ``E_name`` (replaces ``E_str``)
```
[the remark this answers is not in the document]
ScottSoren: For code that should represent something a scientist could conceivably write on paper or on the blackboard, I want `ixdat` to always stay as close as possible to what they'd write. I'd never write "e" for potential or "i" for current (I have seen the latter though). I guess "E" and "I" carry over from the biologic names for the raw data. Calling normalized current "j"  (lowercase) is pretty standard, I think. "v" for calibrated potential might be less standardized (E, V, and U are maybe more common) but capital "V" clashes with the symbol for the unit Volt. Hm... now I kind of regret not calling it "U". what do you think? It could be "U" and "J" for calibrated potential and current, to keep them all uppercase. What do you think? Now's probably the last chance to change them, so it should be something we agree on.

NEXT_CHANGES.rst
```diff
@@ -0,0 +1,102 @@
...
+
+API Changes
+-----------
+
+ixdat.measurement
+^^^^^^^^^^^^^^^^^
+
+- Generalization of multiple calibrations:
+
+  ``Measurement.calibration`` is deprecated in favor of ``Measurement.calibration_list``.
```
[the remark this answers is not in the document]
ScottSoren: It's a getter and a setter... but I think one thing that deprecated. Unless you have a better way to phrase this, I'll leave it be. I expect there will be a development on calibration handling during ixdat v0.2.x, so we can come back to it.

NEXT_CHANGES.rst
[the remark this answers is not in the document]
AnnaWiniwarter: I like "U" much more than "v". Didn't you use "U" also in the figure captions in EC_MS?

NEXT_CHANGES.rst
```diff
@@ -0,0 +1,102 @@
...
+
+- Axes order: Two-panel figures with shared x-axis always return a list of axes in the order
+  ``axes=[top_left, bottom_left]`` in the case of one y-axis of each panel, or
+  ``[top_left, bottom_left, top_right, bottom_right]`` in the case of an additional y-axis
+  in either panel. Note that ``axes[2]`` or ``axes[3]`` might be ``None``. For example,
+  ``axes = ECMSPlotter.plot_measurement()`` by default results in ``axes[0]`` being the
+  MS data, ``axes[1]`` electrode potential, ``axes[2]=None`` and ``axes[3]`` being the
+  electrode current.
+
+- Renamed key-word arguments in EC, EC-MS, and SEC plotting functions:
```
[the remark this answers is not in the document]
ScottSoren: replaced.

NEXT_CHANGES.rst
```diff
@@ -0,0 +1,102 @@
...
+- Reliable construction of `selector` and `cycle` counters. This is done via the
+  ``selector_name`` and ``_series_constructors`` class attributes, which can be
+  customized for every ``Measurement`` subclass.
+
+  This solves `#14 <https://github.com/ixdat/ixdat/issues/14>`_
+
+- Carrying calibrations over through measurement combination (hyphenation or appending
+  with the `+` operator) or transformation (through copying and technique-changing
+  functions like `ECMeasurement.as_cv()`. This is done via replacing the use of a single
+  ``Measurement.calibration`` with a ``calibration_list`` which can be appended,
```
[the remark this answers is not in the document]
ScottSoren: You did review the code that does it ;)

docs/source/developing.rst
```diff
@@ -15,15 +15,15 @@ Git and Github
 The source code for ixdat (and this documentation) lives at:
 https://github.com/ixdat/ixdat
 
-Note, we are currently compiling from the
-`[user_ready] <https://github.com/ixdat/ixdat/tree/user_ready>`_
-branch, not the master branch.
-
 To develop ixdat, you will need to use git and github. This means
 
 - `install git <https://git-scm.com/downloads>`_. Git bash is strongly recommended for Windows
```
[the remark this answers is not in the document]
ScottSoren: the word strongly is removed.

NEXT_CHANGES.rst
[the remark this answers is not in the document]
AnnaWiniwarter: oh, and yes, I guess upper case "J" would make more sense for consistency. though the lower case letter is what one would use for current density on the blackbord, isn't it?

docs/source/developing.rst
```diff
@@ -40,14 +40,18 @@ To develop ixdat, you will need to use git and github. This means
...
-    git checkout branch_to_work_from
+    git switch main
 
-- Make a branch using::
+- Make and switch to a branch on which to develop your feature::
 
-    git branch my_branch_name
-    git checkout my_branch_name
+    git branch my_feature_branch_name
+    git switch my_feature_branch_name
```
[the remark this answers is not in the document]
ScottSoren: The line above it creates the branch.
But that's nice. I'll combine the two commands.

src/ixdat/measurements.py
ScottSoren: Good point. I hadn't thought of `aliases` as static, but I hadn't thought of it as  necessarily dynamic either. No reason to make it dynamic when there's another option. simple better than complex better than complicated, and `cached_series` is clearly already dynamic.
I've changed it to your suggestion.

NEXT_CHANGES.rst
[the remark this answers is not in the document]
ScottSoren: Yeah, typically ``v_name`` is "U vs RHE / [V]"   (same in EC_MS though it was called ``V_str``)
So... yeah lets do "J_name" and "U_name" then. Tribalism for the win. Real electrochemists call potential "U", haha ;) 

On it. Bleh, more refactoring.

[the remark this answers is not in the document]
ScottSoren: was just reminded when opening ec.py to start that ``j_name`` is typically "J / [mA cm$^{-2}$]", haha. 
"E", "I", "U", and "J" are then least confusing.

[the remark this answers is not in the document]
ScottSoren: Oh, but there's more:
``ec_meas.v`` and ``ec_meas.j`` returns the potential and current. Should that be ``ec_meas.U`` and ``ec_meas.J`` ?
And aren't cyclic voltammograms and linear sweeps usually plotted as "J-V curves"?

ScottSoren:
Hi @KennethNielsen and @AnnaWiniwarter , 

I've implemented the changes requested on this PR. This includes describing "mol@mass" and giving an example of ``add_calibration()`` useage in NEXT_CHANGES. This also includes, renaming "v_name" to "U_name" and "j_name" to "J_name" to match what Anna and I feel most natural on the blackboard, keep all the raw and calibrated current and potential shortcuts uppercase, and not clash with the useage of "v" to mean general "value" in ixdat. 

Huge thanks to Anna for testing. I had tested against the usual suspects (tutorials and article repositories) but hadn't found all the bugs. As briefly discussed with Anna on slack, this probably means two more PR's tomorrow before ixdat v0.2.0 release:

- back-compatability with .csv files exported by ixdat v0.1.x
- debugging something about calibration carryover

Let me know if this one is in your eyes good to merge now.

NEXT_CHANGES.rst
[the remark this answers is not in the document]
KennethNielsen: I would suggest something along the lines of "The single calibration gettter and better ... has been deprecated in favor of the ... getter, which will return the current list of calibrations and ... which is used to add a calibration to that list." possibly expanded with "This change is to accommodate the idea of multiple calibrations for a measurement which is used to ..."

Written on my phone, so I will forego typing out the property names 😁

NEXT_CHANGES.rst
```diff
@@ -0,0 +1,102 @@
...
+- Renamed properties in ``ECMeasurement`` and inheriting classes:
+
+  - ``v_name`` (replaces ``V_str``)
+  - ``j_name`` (replaces ``J_str``)
+  - ``E_name`` (replaces ``E_str``)
+  - ``I_name`` (replaces ``I_str``)
+
+  The old property names are deprecated.
+
+- Renamed keyward in ``MSMeasurement.plot_flux()``:
```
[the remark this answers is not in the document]
KennethNielsen: it appears to still be spelled wrong

KennethNielsen (approved):
I've left a few comments on discussions and unresolved a single discussion  around a typo. They are all minor and so once you have had a look at the comments I consider this good to go.

NEXT_CHANGES.rst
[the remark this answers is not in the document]
AnnaWiniwarter: Yes, I guess it should be `ec_meas.U` and `ec.meas.J`, even though when writing with this syntax it feels more like code and less like electrochemistry, so the capitalization seems a bit off. and yes, they are called J-V curves, but I with `U_name` being "U vs RHE / [V]" I think having the U there is the least confusing, as it draws attention to the naming convention in the plot. if someone doesn't like the U there, they are warned and can change it to V manually.

AnnaWiniwarter:
I think it would be great to also have the new syntax for an EC calibration in NEXT_CHANGES, unless you think it shouldn't go in there because `ec_data.calibrate(E_vs_RHE=1, A_el=0.5) ` still works. Else I think it's fine to merge.

NEXT_CHANGES.rst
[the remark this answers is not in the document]
ScottSoren: Done. and more.

ScottSoren:
Alright, @AnnaWiniwarter , I think I have addressed all your concerns here. What was done since last review:

- Renamed `v_name` to `U_name`, `j_name` to `J_name`, etc, updated also where I could find it in documentation, deprecation decorators, and variable names within functions. 
- Generalized the `calibrate` method up from `ECMeasurement` to `Measurement` so that you can use `calibrate` for any type of measurement as a shortcut (rather than `add_calibration`). 
- Described and gave examples of `calibrate` in `NEXT_CHANGES.rst`
- Described `mol@mass` notation with an example in `NEXT_CHANGES.rst` (should go in ms documentation too)
- Described the workflow of using  `CyclicVoltammogram` with an example in [docs/techniques/electrochemistry](https://github.com/ixdat/ixdat/blob/lets_deprecate/docs/source/technique_docs/electrochemistry.rst)
- Ensured that future .csv exports state the ixdat version in the top line
- Added documentation of how to read old .csv's in NEXT_CHANGES.rst

Let me know if all is good! If so I think this is indeed the last PR to v0.2.0 as hoped :) 

@KennethNielsen , since there were some non-trivial changes resulting from Anna's testing, you may also want to give another quick glance. 

Looking forward to merging this!

AnnaWiniwarter:
Looks great! The only comment I have is that in the documentation you now recommend to use redefine cycle, but you don't explain why cycles are messed up if you don't use "redefine cycle" like you wrote me earlier:  "Before you've used this, the` counter cycle` is the "cycle number" column of the imported data. If you use a loop, this repeats itself, so you can get chucks of disconnected data when selecting for it. If you run `redefine_cycle` , the` counter cycle` is redefined so each cycle has a unique number." 
But with the recommendation and the example you give it should be fine to just leave it like it is, since now there is a guide stating that it's a good idea to use it.

NEXT_CHANGES.rst
```diff
@@ -0,0 +1,200 @@
...
+    >>> my_ecms_meas.calibrate(RE_vs_RHE=0.656)   # overshadows the first RE_vs_RHE
+    >>> my_ecms_meas.plot(mol_list=["H2", "O2"])
+
+  3. You can calibrate all at once.
+
+    >>> my_ecms_meas = my_ec_meas + my_ms_meas
+    >>> my_ecms_meas.calibrate(
+    >>>     RE_vs_RHE=0.715, A_el=0.196, ms_cal_results=[my_H2_at_M2, my_O2_at_M32]
+    >>> )   # note that all of these are keyword arguments to ECMSCalibration.
+    >>> my_ecms_meas.plot(mol_list=["H2", "O2"])
```
KennethNielsen, line 69: Looks great and I understand the mechanism better now. In respect to my question the answer then is, if I understand it correct, that concatenating measurements with different calibrations makes no sense, so we will simple use the one that ends up the latest for any given parameter

NEXT_CHANGES.rst
```diff
@@ -0,0 +1,200 @@
...
+
+- Renamed keyword argument in MS and EC-MS plotting functions:
+
+  - ``remove_background`` (replaces ``removebackground``)
+
+  The old argument name is deprecated
+
+ixdat.readers
+^^^^^^^^^^^^^
+- ixdat-eported .csv compatability
```
KennethNielsen, line 138: some typo here in eported

NEXT_CHANGES.rst
```diff
@@ -0,0 +1,200 @@
...
+  of the technique class. For example:
+
+    >>> form ixdat.techniques import ECMeasurement
+    >>> ECMeasurement.essential_series_names
+    {'t', 'raw_potential', 'raw_current'}
+
+  For even earlier .csv files (exported by ixdat version <0.1.5), you will also need to
+  specify the technique.
+
+  Starting in ixdat v0.2.0, ixdat-exported files will have the ixdat version in the header.
```
KennethNielsen, line 168: Good idea

docs/source/developing.rst
```diff
@@ -40,14 +41,17 @@ To develop ixdat, you will need to use git and github. This means
 
     pip install --upgrade ixdat
 
-- Check out the branch you want to work from. Note, this is also how to *use* a feature that is under development.::
+- Switch to the branch you want to work from. Note, this is also how to *use* a feature that is under development.::
+
+    git switch branch_to_work_from
 
-    git checkout branch_to_work_from
+  Usually it makes most sense to work from the main branch::
```
KennethNielsen, line 48: Maybe the formulation should be "Usually ... to **start** work from ..."

KennethNielsen: so as to not cause confusion below when you fire up the feature branch

docs/source/technique_docs/electrochemistry.rst
```diff
@@ -4,22 +4,22 @@ Electrochemistry
 ================
 
 The main TechniqueMeasurement class for electrochemistry is the ``ECMeasurement``.
-Subclasses of ``ECMeasurement`` include ``CyclicVoltammagram`` and ``CyclicVoltammagramDiff``.
+Subclasses of ``ECMeasurement`` include ``CyclicVoltammogram`` and ``CyclicVoltammogramDiff``.
```
KennethNielsen, line 7: Don't remember if I saw a deprecation on the mis-spelled Diff as well

docs/source/technique_docs/electrochemistry.rst
```diff
@@ -29,6 +29,46 @@ the following Tutorials:
 
 - `Analyzing cyclic voltammagrams <https://github.com/ixdat/tutorials/blob/main/simple_ec_analysis/difference_between_two_cvs.ipynb>`_
 
+
+``CyclicVoltammogram`` inherits from ``ECMeasurement`` and adds useful functionality for
+cyclic voltammotry analysis. This includes:
```
KennethNielsen, line 34: "..mmEtry andlysis..."

docs/source/technique_docs/electrochemistry.rst
```diff
@@ -29,6 +29,46 @@ the following Tutorials:
...
+
+  >>> from ixdat import Measurement
+  >>> my_ec_meas = Measurement.read("my_data.mpt", reader="biologic")
+  >>> my_cv = my_ec_meas.as_cv()
+  >>> my_cv.redefine_cycle(start_potential=0.2, redox=True)
+  >>> my_cv.plot_measurement(J_name="cycle")   # co-plots potential and cycle counter
+  >>> ax = my_cv[2].plot(color="b", label="earlier")
+  >>> my_cv[50].plot(ax=ax, color="r", label="later")
+  >>> ax.legend()
+  >>> ax.get_figure().savefig("two_cycles.png")
```
KennethNielsen, line 66: that is beautiful API wise

src/ixdat/exporters/csv_exporter.py
```diff
@@ -84,6 +85,8 @@ def prepare_header_and_data(self, measurement, v_list, tspan):
             s_list.append(v_name)
 
         header_lines = []
+        ixdat_version_line = f"ixdat version = {__version__}\n"
```
KennethNielsen, line 88: You can also choose to version the file format, depending on what is most convenient

src/ixdat/plotters/ec_plotter.py
```diff
@@ -37,18 +42,19 @@ def plot_measurement(
...
-                new twinx()'d axes. axes[0] is for V_str and axes[1] for J_str.
-            V_str (str): DEPRECIATED. now v_name
-            J_str (str): DEPRECIATED. now j_name
-            v_color (str): The color to plot v_name. Defaults to black.
-            j_color (str): The color to plot j_name. Defaults to red.
+            U_color (str): The color to plot U_name. Defaults to black.
+            J_color (str): The color to plot J_name. Defaults to red.
+            V_str (str): DEPRECATED. Use `U_name`.
+            J_str (str): DEPRECATED. Use `J_name`.
+            V_color (str): DEPRECATED. Use `U_color`.
```
KennethNielsen, line 57: Is that "Use `v_color`"? At least that's what's in the decorator

src/ixdat/plotters/ec_plotter.py
```diff
@@ -58,39 +64,45 @@ def plot_measurement(
...
-                "DEPRECIATION WARNING! V_str has been renamed v_name and J_str has "
-                "been renamed j_name. Get it right next time."
-            )
-        v_name = v_name or V_str or measurement.v_name
+
+        # apply deprecated arguments (the user will get a warning):
+        U_name = U_name or V_str
+        J_name = J_name or J_str
+        U_color = U_color or V_color
+        J_color = J_color or J_color
```
KennethNielsen, line 72: Is there missing an combination with old arg here for `v_color`?

KennethNielsen (approved):
I noticed a few more typos, which should of course be fixed, but I will mark as approved, so you won't have to wait for me before merge.

NEXT_CHANGES.rst
ScottSoren: exported.

docs/source/developing.rst
ScottSoren: Usually it makes most sense to start work by branching from the main branch

docs/source/technique_docs/electrochemistry.rst
ScottSoren: that word can go spell itself.

src/ixdat/exporters/csv_exporter.py
ScottSoren: this is most convenient.

src/ixdat/plotters/ec_plotter.py
ScottSoren: `v_color` was a rejected intermediate between `V_color` and `U_color` which never saw PyPI. (Same with `V_name`.) Decorator fixed.

ScottSoren:
Thanks @AnnaWiniwarter and @KennethNielsen for your reviews and tests of this! 
Merging now to main. I'll then change the version to 0.2.0 (removing the "dev"), copy the contest of NEXT_CHANGES.rst to CHANGES.rst, and then upload the main branche's docs to ixdat.readthedocs.org and its packaged code to pypi.org/project/ixdat

PR #51, opened 2022-03-21
change removebackground to remove_background to avoid TypeError in pl…

AlexanderKrabbe:
TypeError when running ms_plotter.plot_vs()  

apparently plot_vs uses "removebackground" and grab_flux uses "remove_background".

I guess it is to preserve backwards compatibility not to change the name of the input in plot_vs and in ecms_plotter line 217 

I cannot find other places where removebackground is used instead of remove_background.
[https://github.com/ixdat/ixdat/search?q=removebackground](url)


ScottSoren:
Hey Alexander. Nice debug. Will merge. 
Be ready: In #50 , I've changed it to `remove_background` everywhere with `DeprecationWarnings` if a user tries to use `removebackground` instead.

PR #55, opened 2022-03-25
Add temperature dependence of dynamic viscocities

AlexanderKrabbe:


ScottSoren:
Alexander has come accross something very interesting for MS quantification!
@KennethNielsen and @AnnaWiniwarter , do you think that this implementation is good for T-dep. of dynamic viscosities?

AnnaWiniwarter (approved):
Looks like a decent solution to me.

src/ixdat/techniques/ms.py
```diff
@@ -543,7 +543,9 @@ def calc_n_dot_0(self, gas=None, w_cap=None, h_cap=None, l_cap=None, T=None, p=N
             p = self.p
 
         pi = np.pi
-        eta = DYNAMIC_VISCOSITIES[gas]  # dynamic viscosity in [Pa*s]
+        _eta_v = list(DYNAMIC_VISCOSITIES[gas].values())  # list of eta(T) for 'gas'
+        _eta_T = list(DYNAMIC_VISCOSITIES[gas].keys())  # list of Ts for 'gas'
```
KennethNielsen, line 547: Hmm. This spawned in interesting thought, which I think I have settled for myself once before but forgotten in the mean time, which is if keys and values are guarantied to come out in the same order and if that order is insertion order (which I think we need to the interpolation). The answer to the first seems to be "yes" no-matter the Python version I have looked at, but the answer to the second seems to be, that it depends on the Python version. Have a look at this link: https://docs.python.org/3.7/library/stdtypes.html#dictionary-view-objects and witch between Python 3.6 (which we still support) and Python 3.7 and notice how the comment in the code example changes from:
```python
>>> # keys and values are iterated over in the same order
```
to 
```python
>>> # keys and values are iterated over in the same order (insertion order)
```

To enforce it you can do something like:
```python
from operator import itemgetter
...
_sorted_eta_pairs = sorted(DYNAMIC_VISCOSITIES[gas].items(), key=itemgetter(0))  # Sorted sequence of (T, eta(T)) pairs for 'gas'
_eta_T, _eta_v = zip(*_sorted_eta_pairs)  # tuple of T's and corresponding tuple of eta(T)'s
```

src/ixdat/techniques/ms.py
```diff
@@ -543,7 +543,9 @@ def calc_n_dot_0(self, gas=None, w_cap=None, h_cap=None, l_cap=None, T=None, p=N
...
+        eta = np.interp(T, _eta_T, _eta_v)  # dynamic viscosity of gas at T in [Pa*s]
```
KennethNielsen, line 548: So I wondered what would happen if the T was out of range of the interpolation input, turns out `np.iterp` coerces to the value at the extremes, `_eta_v[0]` or `_eta_v[-1]`. In order to let the user know that they are going into unchartered waters, I think this is the right sort of thing for a Python warning. It might look something like this (using the already sorted list of T's:
```python
import warnings
...
# On the already sorted list of T
if T < _eta_T[0] or T > _eta_T[-1]:
    warnings.warn(f"Insufficient data in constants.py to appropriately estimate the dynamic viscosity for {gas} at temperature: {T}K")
```

src/ixdat/constants.py
```diff
@@ -26,15 +26,105 @@
...
+        293.15: 0.88e-05,
+        323.15: 0.94e-05,
+        373.15: 1.04e-05,
+        473.15: 1.21e-05,
+        573.15: 1.37e-05,
+        673.15: 1.53e-05,
+        773.15: 1.69e-05,
+        873.15: 1.84e-05,
+    },
+}  # in [Pa*s] engineeringtoolbox.com
```
KennethNielsen, line 126: Great with the sourcing :) Can you maybe move this comment up front and mention the key units

KennethNielsen (changes requested):
Hi Alenxader

This looks great. I straight forward solution cover this problem.
I have a few comments, mostly covering weird corner cases. Please have a look and let me know what you think.

Regards Kenneth

src/ixdat/techniques/ms.py
AlexanderKrabbe: What do you think of alternatively to make the dynamic "constants" a list of list in the dictionary of gasses? 
I represented it as a dictionary in a dictionary for easy read of temperature / eta pairs.

but in ..CONSTANTS it could be changed to
```
DYNAMIC_VISCOSITIES = {
"gas" : [ 
        [T1, eta1],
        [T2, eta2],
    ]
}
```

Then the order is preserved when interpreting

`eta =np.interp(T, DYNAMIC_VISCOSITIES['gas'][:,0], DYNAMIC_VISCOSITIES['gas'][:,1]`

and the easy readabilities is preserved compared to my initial thought of lists in the constants.

src/ixdat/techniques/ms.py
AlexanderKrabbe: Yes exactly, and I havent used warnings before so I thought of using a try except but that seemed wrong and I decided not to. 

I will try and implement a warning.

src/ixdat/techniques/ms.py
AlexanderKrabbe: However, this would require numpy to be imported in the constants, which I do not know if that is bad practice or not

KennethNielsen: I don't consider it a problem to import numpy in constants and I think that is a fine solution on all accounts.

src/ixdat/constants.py
AlexanderKrabbe: I hope I have put the comment the correct place.

KennethNielsen: :D it wasn't that I was commenting on your comment placement, it was just that when the dict was small it didn't really matter where it was, but now that it is so many lines, it makes more sense with it before. In any case, the units in the temperature was the important stuff.

src/ixdat/techniques/ms.py
KennethNielsen: Look good

src/ixdat/techniques/ms.py
```diff
@@ -543,7 +544,17 @@ def calc_n_dot_0(self, gas=None, w_cap=None, h_cap=None, l_cap=None, T=None, p=N
             p = self.p
 
         pi = np.pi
-        eta = DYNAMIC_VISCOSITIES[gas]  # dynamic viscosity in [Pa*s]
+        if T < DYNAMIC_VISCOSITIES[gas][0, 0] or T > DYNAMIC_VISCOSITIES[gas][-1, 0]:
+            warnings.warn(
+                f"Insufficient data in constants.py to appropriately estimate\
```
KennethNielsen, line 549: Just a tiny nit-pick. This line continuation doesn't work quite the way you expected I suspect. If you put the continuation inside the string, then the string will contain the space before the next line also:
```python
>>> print("My very long string\
...        that continues on the next line",
...        "argument 2")
My very long string       that continues on the next line argument 2
```
So, you need to turn both into strings and continue those:
```python
>>> print("My very long string "\
...       "that continues on the next line",
...       "argument 2")
My very long string that continues on the next line argument 2
```
But then, because you are inside a parenthesis, the continuation of strings happens automatically, so you can drop the backslash:
```python
>>> print("My very long string "
...       "that continues on the next line",
...       "argument 2")
My very long string that continues on the next line argument 2
```

KennethNielsen (changes requested):
Looks great. Just a tiny nit-pick on line continuation and then good to go.

Also, if you branched your feature branch of at a point in time where Scott had alread created the `NEXT_CHANGES.rst` file at the root of the repository, can you make a small addition to that, that describes this improvement?

ScottSoren (approved):
Beautiful addition! Thanks for this. It looks like anyone doing all their measurements at standard temperature will never notice the change, but that people doing temperature-dependent measurements will have a much more powerful tool after this change. I agree with Kenneth that the numpy arrays are better than ditionaries for this, so that was a nice improvement. 
Documentation is missing in NEXT_CHANGES.rst, but I'm happy to add that as I know you are busy.

AlexanderKrabbe:
thanks. for some reason my git was messed up and I could not pull or rebase in a proper way :-/

PR #60, opened 2022-03-29
Export data density

ScottSoren:
#57 should merge first.

This allows the user to reduce the size on disk of an exported masurement by specifying a `time_step` in the export function.
Example result of such an export here: 

https://github.com/ScottSoren/Huang2021/blob/main/data/fig3ab.csv

Notice the "projected time / [s]" added to "aliases" which enables the Measurement to find it. It's tested only against the Huang2021 repository so far.

One more commit comming with documentation in NEXT_CHANGES.rst


NEXT_CHANGES.rst
```diff
@@ -11,3 +11,54 @@ links to relevant Issues, Discussions, and PR's on github with the following for
...
+
+- ``export()`` functions now take a ``time_step`` argument. It must be used together with
+  a ``tspan``. It is effective in reducing the file size of exported .csv's.
+
+- Renamed keyward argument: ``columns`` replaces ``v_list in exporters (``ECExporter`` in v0.2.0, ``ECMSExporter`` in 0.2.1)
+
+  The old name is deprecated.
+
+- ``ECMSExporter.export()`` functions now take a ``time_step`` argument. It must be used together with
+  a ``tspan``. It is effective in reducing the file size of exported .csv's.
```
AnnaWiniwarter, line 53: Is this a duplicate of line 45&46 (which seems to be more general)

AnnaWiniwarter (approved):
This is a nice idea, and works on the dataset I tested it on (as long as I choose a somewhat reasonable time_step (with 50s time step, cutting cvs into the actual cycles doesn't work anymore, but I think that's just for the lack of information, so I don't see this as a bug).

Out of curiosity though, why does it require that I select a tspan, in order to use time_step? I usually save the whole dataset as csv... (very nice though, I got an error message that provided the info I needed to get it to run)

(as usual, I didn't focus on reading the code, I trust @KennethNielsen will take care of this :)

NEXT_CHANGES.rst
ScottSoren: yes. fixed.

ScottSoren:
Hi @AnnaWiniwarter , I've fixed the RE_vs_RHE=0 thing here, and cleaned up the exported `aliases` mess a bit more. You may want to give it another look?

AnnaWiniwarter:
It's working for me now (the RHE problem)! Thanks for fixing. :)

src/ixdat/exporters/ecms_exporter.py
```diff
@@ -1,36 +1,67 @@
...
+        columns=None,
         tspan=None,
         mass_list=None,
         mol_list=None,
+        time_step=None,
     ):
-        if not v_list:
+        """Export a given measurement to a specified file.
+
+        To improve flexibility with inheritance, this method allocates its work to:
```
KennethNielsen, line 32: Good, but, as I understand it, that probably means that these methods fall through to `CSVExporter` via `ECExporter`, which is now what it inherits from. That is great, and all as intended, but I'm not quite sure that I think the comment will help a less programming inclined reader will understand that from the comment, but I may just be reading it wrong. What about something like:
```text
This method delegates the majority of the export work, via inheritance, to:
 - ...
 - ...
 - ...
```

KennethNielsen: In any case, I would suggest "delegates" instead of "allocates" in your original comment

src/ixdat/exporters/ecms_exporter.py
```diff
@@ -1,36 +1,67 @@
...
+        - CSVExporter.prepare_header_and_data()
+        - CSVExporter.write_header()
+        - CSVExporter.write_data()
+
+        Args:
```
KennethNielsen, line 37: Just as a general comment, it is a lot easier during review to ensure completeness of the Args list, if it is sorted the same way as the arguments :D

src/ixdat/exporters/ecms_exporter.py
```diff
@@ -1,36 +1,67 @@
...
+            measurement (Measurement): The measurement to export.
+                Defaults to self.measurement.
+                TODO: remove this kwarg. See conversation here:
+                   https://github.com/ixdat/ixdat/pull/30/files#r810926968
+            path_to_file (Path): The path to the file to write. If it has no suffix,
+                a .csv suffix is appended. Defaults to f"{measurement.name}.csv"
+            columns (list of str): The names of the data series to include. Defaults in
```
KennethNielsen, line 44: I think this one is a little generic. If you want to mention what it defaults to, you should probably (if it is possible) mention what it defaults to _in the case of an ECMS exporter_

src/ixdat/exporters/ecms_exporter.py
```diff
@@ -1,36 +1,67 @@
...
+                CSVExporter to all VSeries and TSeries in the measurement. This default
+                may be overwritten in inheriting exporters.
+            v_list: DEPRECATED. Use `columns` instead.
+            tspan (timespan): The timespan to include in the file, defaults to all of it
+            time_step (float): Optional. The time spacing between data points. Can be
```
KennethNielsen, line 49: Mention here, that it requires a tspan argument?

src/ixdat/exporters/sec_exporter.py
```diff
@@ -40,9 +39,7 @@ def default_export_columns(self):
         )
         return v_list
 
-    aliases = ECExporter.aliases
-
-    def prepare_header_and_data(self, measurement, v_list, tspan):
+    def prepare_header_and_data(self, measurement, v_list, tspan, time_step=None):
         """Do the standard ixdat csv export header preparation, plus SEC stuff.
```
KennethNielsen, line 43: Should this docstring refer to the ECExporter methods for explanation or args?

src/ixdat/plotters/ms_plotter.py
```diff
@@ -165,7 +165,7 @@ def plot_vs(
         unit=None,
         logplot=True,
         legend=True,
-        **kwargs,
+        **plot_kwargs,
```
KennethNielsen, line 168: This also looks like a change I have seen before :|

KennethNielsen: Or maybe it is just in a new plotter

KennethNielsen (changes requested):
Looks good. There are a few typos and such, but nothing serious.

One thing that I noted, is that when I click the like to the exported file, github is helpfully letting me know that there might be an issue with the CSV format at line 8, which is an exporter dict as JSON, most likely because there are lots of , in there. This will at the very least cause problems if you try to import the file elsewhere than in ixdat, which handles the header differently than as pure CSV. I don't necessarily think it should be handled here, but if you decide not to, let me know and I will make an issue for it.

NEXT_CHANGES.rst
```diff
@@ -11,3 +11,53 @@ links to relevant Issues, Discussions, and PR's on github with the following for
...
+  colors both the U-J curve in the lower panel and all the signals in the top panel,
+  so best to use with together with a ``mass_list`` or ``mol_list`` of length 1. (as of 0.2.1)
+
+exporters
+^^^^^^^^^
+
+- ``export()`` functions now take a ``time_step`` argument. It must be used together with
+  a ``tspan``. It is effective in reducing the file size of exported .csv's.
+
+- Renamed keyward argument: ``columns`` replaces ``v_list in exporters (``ECExporter`` in v0.2.0, ``ECMSExporter`` in 0.2.1)
```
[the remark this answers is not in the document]
ScottSoren: nice catch!

requirements-dev.txt
```diff
@@ -1,4 +1,4 @@
-black==22.1.0
+black==22.3.0
```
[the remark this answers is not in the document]
ScottSoren: That will be no problem.

src/ixdat/exporters/csv_exporter.py
```diff
@@ -48,31 +61,63 @@ def export(self, path_to_file=None, measurement=None, columns=None, tspan=None):
...
+        return aliases
+
+    def prepare_header_and_data(self, measurement, columns, tspan, time_step=None):
         """Prepare self.header_lines to include metadata and value-time pairs
 
         Args:
             measurement (Measurement): The measurement being exported
-            v_list (list of str): The names of the ValueSeries to include
+            columns (list of str): The names of the ValueSeries to include
             tspan (timespan): The timespan of the data to include in the export
```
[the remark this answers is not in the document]
ScottSoren: Nice. Done.

src/ixdat/exporters/csv_exporter.py
```diff
@@ -48,31 +61,63 @@ def export(self, path_to_file=None, measurement=None, columns=None, tspan=None):
...
         """
         columns_data = {}
         # list of the value names to export:
-        v_list = v_list or self.default_export_columns or list(measurement.value_names)
+        columns = columns or self.default_export_columns or list(measurement.value_names)
+        self.columns = columns
```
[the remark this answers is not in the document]
ScottSoren: Okay. But that makes the line more than 89 chars :(

src/ixdat/exporters/csv_exporter.py
```diff
@@ -48,31 +61,63 @@ def export(self, path_to_file=None, measurement=None, columns=None, tspan=None):
...
         s_list = []  # list of the series names to export.
         # s_list will also include names of TimeSeries.
 
+        if time_step:
+            if not tspan:
+                raise TypeError(
```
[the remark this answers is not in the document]
ScottSoren: yes and yes.

src/ixdat/exporters/csv_exporter.py
```diff
@@ -48,31 +61,63 @@ def export(self, path_to_file=None, measurement=None, columns=None, tspan=None):
...
+                    "A `time_spacing` can only be specified for an export if a `tspan` "
+                    "is also specified."
+                )
+            t = np.arange(start=tspan[0], stop=tspan[-1], step=time_step)
```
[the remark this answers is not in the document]
ScottSoren: I'll answer here, also for @AnnaWiniwarter : 

In general, a dataset of a hyphenated measurement technique can have a much longer timespan in some of its value series than others. Presently, the interpolation (in `grab_for_t`) uses the default `vp[0]` where `t<tp[0]` and `vp[-1]`  where `t>tp[-1]`, i.e. it repeats the endpoints. That results in some dumb-looking and space-wasting .csv's. Without requiring the user to think about what timespan they're exporting, they can in fact get **larger** files as a result of specifying a `time_step` .
Not sure where this belongs in documentation. I've added the requirement in the kwarg documentation.

src/ixdat/exporters/ecms_exporter.py
ScottSoren: both orders changed to be in sync

src/ixdat/exporters/sec_exporter.py
ScottSoren: yes.

src/ixdat/plotters/ms_plotter.py
ScottSoren: There's so many of these sneaky little kwargs that escaped a refactor.

ScottSoren:
@KennethNielsen , waiting for the green light :)

KennethNielsen (approved):
Looks good to go :)

PR #66, opened 2022-04-04
debug for ecms_calibration_curve

ScottSoren:
A couple small things came up for debugging while attempting to use `ECMSMeasurement.ecms_calibration_curve()`. They are described in NEXT_CHANGES.rst


KennethNielsen:
> ... They are described in NEXT_CHANGES.rst

It's a nice side-effect, how that is kind of doubling as description here on the PR.

NEXT_CHANGES.rst
```diff
@@ -13,3 +13,38 @@ links to relevant Issues, Discussions, and PR's on github with the following for
...
 
+For ixdat 0.2.2
+===============
+
+API Changes
+-----------
+
+- ``ECMSMeasurement.ecms_calibration_curve()`` no longer returns axes by default.
+  Instead it only returns the calculated ``MSCalResult``. The axes on which the result
+  is plot can be returned by setting the keyword argument ``return_axes=True``
```
KennethNielsen, line 24: ... plotted...

NEXT_CHANGES.rst
```diff
@@ -13,3 +13,38 @@ links to relevant Issues, Discussions, and PR's on github with the following for
...
+  changes in specified columns (common ones include "file_number", "cycle_number", etc).
+  Especially useful for compound techniques with biologic potentiostats.
+
+Bug Fixes
+---------
+
+- ``ECMSMeasurement.ecms_calibration_curve()`` now plots on the correct axes when given
+  ``axes_measurement``.
+
+- Adding calibration data to a measurement with ``calibrate()`` now clears the cache.
```
KennethNielsen, line 37: Uhhh, that's nasty 😮

src/ixdat/measurements.py
```diff
@@ -926,6 +924,28 @@ def _build_selector_series(
...
+    def rebuild_selector(self, selector_string=None, columns=None, extra_columns=None):
+        """Build a new selector series for the measurement and cache it.
+
+        This can be useful if a user wants to change how their measurement counts
+        sections (for example, only count sections when technique or file number changes)
+
+        Args:
+            selector_string (str): The name to use for the selector series
+            columns (list): The list of demarcation series. The demarcation series have
+                to have the same tseries, which should be the one pointed to by the
```
KennethNielsen, line 936: I can't see this enforced, but I don't know if it should be or if I missed it.

KennethNielsen: Ah wait, maybe that's one of the warnings above.

src/ixdat/readers/zilien.py
```diff
@@ -35,10 +35,16 @@ def read(self, path_to_file, cls=None, name=None, **kwargs):
 
         from EC_MS import Zilien_Dataset
 
-        if cls is Measurement:
-            cls = ECMSMeasurement
-
-        if "technique" not in kwargs:
+        if "technique" in kwargs:
```
KennethNielsen, line 38: Nice. I'm actually re-writing the Zilien parser at this moment, but I will make sure to include this part.

src/ixdat/techniques/ec_ms.py
```diff
@@ -133,18 +134,20 @@ def ecms_calibration_curve(
                 Defaults to a new axis.
             axes_measurement (list of Axes): The EC-MS plot axes to highlight the
                 ms_calibration on. Defaults to None.
+            return_axes (bool): Whether to return the axes together with the MSCalResult.
+                Defaults to False.
```
KennethNielsen, line 138: Should this mention the three-part return value option, where setting this to True will return 2 objects, but three if axis_measurement is also given?

KennethNielsen (changes requested):
Looks good to me, only a few minor comments.

NEXT_CHANGES.rst
ScottSoren: super satisfying to fix though :)

src/ixdat/techniques/ec_ms.py
ScottSoren: Actually, there is no need to return axes_measurement, since the user of this function must already have them. So the function only has one or two returns based on the value of `return_axes`.

ScottSoren:
Awaiting your green light!

src/ixdat/techniques/ec_ms.py
KennethNielsen: I still don't think this does what the docstring says it should. As of now the docstring says:
```text
            ax (Axis): The axis on which to plot the ms_calibration curve result.
                Defaults to a new axis.
           ...
           return_axes (bool): Whether to return the axis on which the calibration is
                plotted together with the MSCalResult. Defaults to False.
        Return MSCalResult(, Axis): The result of the ms_calibration (and calibration
            curve axis if requested)
```
and the code says:
```python
        if ax and return_axes:
            if axes_measurement:
                return cal, ax, axes_measurement
            return cal, ax
```
but, there is no reason to test for ax, it will always be there. That is also the case further up. If this is meant as a way to allow toggling the plotting of the fit on/off, then the `ax` argument should probably default to None and the "new" option documented. But as it is now, the only way that ex will ever not be True is if the user explicitly provides an argument that evals to False, which since that option isn't documented, seems fairly exotic.

If, what you meant with the `ax` option merely was, that the user can choose **where** the fit is plotted, not if it is, then both tests in this method for `if ax` can be removed.

But back to the return values which was what we were originally talking about. There still exists the 3 value return, suggest it should be:
```python
if return_axis:
    return cal, ax
return ax
```
or if you want to allow for and document not having an ax:
```python
if return_axis and ax:
    return cal, ax
return ax
```

but I suspect that the first option is really what you want.

KennethNielsen:
> Awaiting your green light!

Just one of the comments still not resolved I think. Let me know when you have had a look.

ScottSoren:
In addition to fixing the return conditional you pointed out, this PR now also has an important thing:

- `MSMeasurement.grab_flux()` can take a `MSCalResult` as its first argument. This is now documented in NEXT_CHANGES.rst

KennethNielsen (approved):

PR #69, opened 2022-04-07
Zilien parser

ScottSoren:
So that I have the code on ixdat/ixdat for testing.


KennethNielsen:
@ScottSoren may I ask what it was that you needed here. I seems like a little roundabout workflow, that I think we should see if we can eliminate.

PR #74, opened 2022-05-18
Add tspan selection based on selector for ECMS calibration

AnnaWiniwarter:
Motivated by the broader functionality of the calibration methods in EC_MS, I added an option to `ecms.ecms_calibration_curve` which automatically selects the tspans when given a list of selectors (e.g. cycles or different CP or CA segments). 
https://github.com/ScottSoren/EC_MS/blob/f7c12f9f2180950ca929d6da9ca2767479a2e106/src/EC_MS/Calibration.py

Because of the limitations of the syntax of `.select_values()` and my Python skills, the code is currently pretty ugly. Please have a look and let me know if you have any suggestions for improvement.

### DIFF
--- a/src/ixdat/techniques/ec_ms.py
+++ b/src/ixdat/techniques/ec_ms.py
@@ -118,7 +118,7 @@ class ECMSMeasurement(ECMeasurement, MSMeasurement):
         tspan_list=None,
         selector_list=None,
         selector_name=None,
-        tspan_steady_pulse=0,
+        t_steady_pulse=0,
         tspan_bg=None,
         ax="new",
         axes_measurement=None,
@@ -149,7 +149,7 @@ class ECMSMeasurement(ECMeasurement, MSMeasurement):
         n_list = []
         if not tspan_list:
             tspan_list = self._get_tspan_list(selector_list,
-                                              selector_name, tspan_steady_pulse)
+                                              selector_name, t_steady_pulse)
         for tspan in tspan_list:
             Y = self.integrate_signal(mass, tspan=tspan, tspan_bg=tspan_bg, ax=axis_ms)
             # FIXME: plotting current by giving integrate() an axis doesn't work great.
@@ -187,7 +187,7 @@ class ECMSMeasurement(ECMeasurement, MSMeasurement):
             self,
             selector_list,
             selector_name=None,
-            tspan_steady_pulse=0,
+            t_steady_pulse=0,
     ):
         """
         Generate a t_span list from input of selectors.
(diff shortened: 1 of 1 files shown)


src/ixdat/techniques/ec_ms.py
```diff
@@ -177,6 +183,45 @@ def ecms_calibration_curve(
...
+                                                refer to data['selector']
+            tspan_steady_pulse (float): length of steady state pulse period to integrate
+                                    (will choose the last x seconds)
+
+        Returns tspan_list(list of tspan)
+        """
+        t_index = -1
+        if tspan_steady_pulse == 0:
+            t_index = 0
+        if not selector_name or selector_name == "selector":
```
ScottSoren, line 208: Aah, I see, this is why it was annoying that `select` took kwargs rather than strings to identify the name of the value to be selected for. There is, however, this way around this, using the `**kwargs` trick in reverse:
You can use `self.select_values(**{selector_name: selector_value})`

If you prefer to leave the code as it is, then you should specify in the docstring (here and in `ecms_calibration_curve` that selector_name can only be "selector", "cycle", or "Ns".


An annoying problem with the `**kwargs notation` is that there can be no spaces or wierd characters in `selector_name` for this to work, whereas in general ixdat `DataSeries` can have any string as their name. This should maybe be written in the docstring.
I think that `Measurement.select_value`, at least, should have the option of specifying the `selector_name` as a string rather than with the `**kwargs` trick, so that for it's possible to select for data series that have spaces or wierd characters in their names. I will open an Issue on this for now.

src/ixdat/techniques/ec_ms.py
```diff
@@ -177,6 +183,45 @@ def ecms_calibration_curve(
...
+    ):
+        """
+        Generate a t_span list from input of selectors.
+
+        Args:
+            selector_list (list of selector): selector numbers that define the
+                                            tspans over which data should be integrated
+            selector_name (str): selector name that should be, which by default will
+                                                refer to data['selector']
+            tspan_steady_pulse (float): length of steady state pulse period to integrate
```
ScottSoren, line 200: I think this should just be called `t_steady_pulse`. The use of `tspan` to me indicates it should be a `(t_start, t_end)` tuple.

ScottSoren (commented):
Hi @AnnaWiniwarter ,

This looks great!
I have a couple comments here. One about `**kwarg` trick in reverse to make the code simpler (and an apology that there's not a way around it right now), and an argument name.
In addition to that, there's two more things I need in order to merge this PR:
1. Formatting. Right now it fails flake8, and thus gets the red x with the github tools :(, because some lines are too long (max length is 89 chars).
2. Write what you did in NEXT_CHANGES.rst

src/ixdat/techniques/ec_ms.py
ScottSoren: Issue now opened on this: #77

AnnaWiniwarter: ahh, I wasn't familiar with how to use the **kwargs trick in reverse. I tried with the dict without the stars and it didnt work, so I concluded it doesnt work. That's a much neater solution than mine. :)

AnnaWiniwarter:
I made the changes you suggested, the reverse **kwargs trick is working! :) 
I also made sure that no lines are longer than 89 characters, and I'm using Spyder's automatic linting and code formatting using autopep8, but it still failed the checks. :( Any suggestions what else I need to do to pass?

ScottSoren:
Hey Anna,

This looks good! 

I think it's the indentation in multi-line function calls that triggers the style police. It doesn't fail linting any more, it fails a "check black formatting" step that Kenneth put in. Best just to use black and move on from all formatting discussions ;) . No worries though, lets just get this merged. 

Just needs a few sentences in NEXT_CHANGES.rst. I'll do that this time since it's quick and since I was slow to respond on this PR, but it saves time both when reviewing a PR and when when preparing a release if the person making the PR does it!

AnnaWiniwarter:
thanks for adding the sentences to NEXT_CHANGES. I realised I had forgotten about this after I committed the update, but didnt have time to add it since. Will try to remember next time! :)

PR #81, opened 2022-05-30
Add Github Action to ansure there is changelog entry in NEXT_CHANGES.rst

KennethNielsen:
This PR adds a Github Action that checks whether a PR contains changes to NEXT_CHANGES.rst. If it doesn't, it qualifies as a failed PR check, preventing the PR from begin merged. The check can be disabled for individual PR's if it is evaluated to be necessary by adding the label "skip-changelog" (which will need to be added to the label collection).

### DIFF
--- a/.github/workflows/changelog_checker.yml
+++ b/.github/workflows/changelog_checker.yml
@@ -1,0 +1,19 @@
+name: "Pull Request Workflow"
+on:
+  pull_request:
+    # The specific activity types are listed here to include "labeled" and "unlabeled"
+    # (which are not included by default for the "pull_request" trigger).
+    # This is needed to allow skipping enforcement of the changelog in PRs with specific labels,
+    # as defined in the (optional) "skipLabels" property.
+    types: [opened, synchronize, reopened, ready_for_review, labeled, unlabeled]
+
+jobs:
+  # Enforces the update of a changelog file on every pull request
+  changelog:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: dangoslen/changelog-enforcer@b2a627f5ad95e3610e22143df08656fc1a8ad130
+      with:
+        changeLogPath: 'NEXT_CHANGES.rst'
+        skipLabels: "skip-changelog"
+        missingUpdateErrorMessage: "Please add a changelog entry to NEXT_CHANGES.rst"
--- a/.github/workflows/changelog_checker.yml
+++ b/.github/workflows/changelog_checker.yml
@@ -1,4 +1,4 @@
-name: "Pull Request Workflow"
+name: "Changelog checker"
 on:
   pull_request:
     # The specific activity types are listed here to include "labeled" and "unlabeled"


KennethNielsen:
(Pssssst: I've actually disabled the check for this PR, because I don't think these sort of small tool chain improvements are changelog worthy)

.github/workflows/changelog_checker.yml
```diff
@@ -0,0 +1,19 @@
+name: "Changelog checker"
+on:
+  pull_request:
+    # The specific activity types are listed here to include "labeled" and "unlabeled"
+    # (which are not included by default for the "pull_request" trigger).
+    # This is needed to allow skipping enforcement of the changelog in PRs with specific labels,
+    # as defined in the (optional) "skipLabels" property.
```
ScottSoren, line 7: Why can't the "skipLabels" property be used without including `labeled` and `unlabeled` in types?

.github/workflows/changelog_checker.yml
```diff
@@ -0,0 +1,19 @@
...
+    types: [opened, synchronize, reopened, ready_for_review, labeled, unlabeled]
+
+jobs:
+  # Enforces the update of a changelog file on every pull request
+  changelog:
+    runs-on: ubuntu-latest
```
ScottSoren, line 13: Why *ubuntu*-latest? Anything windows users should keep in mind?

.github/workflows/changelog_checker.yml
```diff
@@ -0,0 +1,19 @@
...
+    steps:
+    - uses: dangoslen/changelog-enforcer@b2a627f5ad95e3610e22143df08656fc1a8ad130
```
ScottSoren, line 15: Is there a link a curious developer could follow to see a description of this action?

ScottSoren (approved):
Great addition! A nice nudge for developers that will save PR reveiwers and new-version-releasers lots of time :) 
I've put a few questions here for my own curiosity, but will go ahead and merge it.

.github/workflows/changelog_checker.yml
KennethNielsen: Right, so this text is actually copied from a template so I hope I'm right. The point I'm pretty sure is, that the test will not be triggered when you **change** the label, unless you make it listen for those event. The regular CI Github Actions are only run on PR creation, open/close and new commits. So, of you were missing the change log entry and added in a new commit, the test would be re-run and pass. But if you were missing it and decided that the PR shouldn't have one and add the label, the tests will not be re-run and the test will remain failed... Unless you listen for the label changes events.

.github/workflows/changelog_checker.yml
KennethNielsen: Ahh, good point. I have inserted here, the same exact commit of the repository that we use at Spectro, because we security audited that one to make sure it isn't leaking code out. But it of course make the link look a little more obscure. It is just a github path, so the link is here: https://github.com/dangoslen/changelog-enforcer and the page for the GitHub action is here: https://github.com/marketplace/actions/changelog-enforcer. Should I add it in a PR or can you just do it directly on main?

PR #86, opened 2022-06-23
Eclab parser tests

matenestor:
A new submodule with [ixdat-large-test-files](https://github.com/ixdat/ixdat-large-test-files) repository is introduced. The PR code is using files from there in order to perform tests of ixdat's ec-lab parser. Tested parsed values are:

- amount of columns
- column names
- column units
- first and last value in a column


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "submodules/ixdat-large-test-files"]
+	path = submodules/ixdat-large-test-files
+	url = git@github.com:ixdat/ixdat-large-test-files.git
```
KennethNielsen, line 3: I think you will need to fix this up, to work of the https endpoint instead of ssh. Since the repo is public it should clone for everyone as https and this will required people to setup ssh, which we don't want to.

tests/functional/test_eclab_parser.py
```diff
@@ -0,0 +1,251 @@
+"""
+The module tests correct parsing of mpt files generated by EC-lab with use of Zilien.
+The dataset with the files exists as a submodule
+
+NOTE: Test of PEIS technique is skipped, because it contains two columns with
```
KennethNielsen, line 5: I suggest a wording along the lines of "not supported along with normal U, I based measurements".

tests/pytest.ini
```diff
@@ -0,0 +1,4 @@
+[pytest]
+markers =
+    submodule_multiple_techniques: test parsing of EC-lab generated files (deselect with '-m "submodule_multiple_techniques"')
```
KennethNielsen, line 3: I think we need to flip these around. A) I don't think the mark needs to be that specific, I would suggest just "submodule" or "extended_tests" and I would suggest that they are deselected unless specifcally enabled. The purpose would be, that any developer should be able to make a PR running the standard tests without all the submodules and then the full suite will be run only before release or selectively by developers needing them.

tasks.py
```diff
@@ -66,15 +66,20 @@ def flake8(context):
 
 
 @task(aliases=["test", "tests"])
-def pytest(context):
+def pytest(context, color="yes", external=False):
```
KennethNielsen, line 69: I'd add some explanation of that argument. Remember that for invoke tasks, the way to add documentation so that it is nicely formatted in `invoke help` is to add it to the decorator.

tests/functional/test_eclab_parser.py
```diff
@@ -0,0 +1,406 @@
+"""
+The module tests correct parsing of a mpt files generated by the EC-lab with
+use of the Zilien. The dataset with the files exists as a submodule
```
KennethNielsen, line 3: maybe add "the submodule 'ixdat-large-test-files'" to prepare for the future, where we might have more than one

tests/functional/test_eclab_parser.py
```diff
@@ -0,0 +1,406 @@
...
+
+DATA_DIR = Path(__file__).parent.parent.parent / "submodules/ixdat-large-test-files/"
+
+"""
+Data to use in the test.
+
+Format is:
+
+filename_with_technique: {
+  "column_name": ("unit", (first_column_value, last_column_value)),
```
KennethNielsen, line 23: Looks great. If one wanted to expand the test a bit one could imagine expanding with number of points or a column content checksum, but I'm not sure if it is necessary.

tests/functional/test_eclab_parser.py
```diff
@@ -0,0 +1,406 @@
...
+        "cycle number": ("", (0.000000000000000e000, 0.000000000000000e000)),
+        "P/W": ("W", (2.2328596e-008, 2.7340908e-008)),
+        "Ewe-Ece/V": ("V", (-2.5963283e-001, -2.2592688e-001)),
+        "loop_number": ("", (0, 1)),  # added by Ixdat
+    },
+}
+
+
+@fixture(scope="module", params=tuple(TEST_DATA.items()))
+@mark.external
```
KennethNielsen, line 303: I don't think you need to or should mark the fixtures, only the tests

tests/functional/test_eclab_parser.py
```diff
@@ -0,0 +1,406 @@
...
+    measurement = measurements_with_data[0]
+    column_names = measurements_with_data[1].keys()
+
+    # test amount of columns
+    assert len(measurement.series_list) == len(column_names)
+
+    # test names of columns
+    for name in column_names:
+        if name not in measurement.series_names:
+            assert (
```
KennethNielsen, line 330: If this is for a clearer error message then great (although I think the error message may be a copy/paste from another test, at least it isn't very informative for this test). Otherwise you can just do:
```python
assert name in measurement.series_name
```
and pytest will show you the value it fails for if it does, which, along with an aptly named test should contain all the information you would need.

tests/functional/test_eclab_parser.py
```diff
@@ -0,0 +1,406 @@
...
+    """
+    measurement = measurements_with_data[0]
+    column_data = measurements_with_data[1]
+
+    for column_name, unit_values in column_data.items():
+        expected_unit = unit_values[0]
+
+        assert (
+            expected_unit == measurement[column_name].unit_name
+        ), f"Parsed value does not match with the value in a column: {column_name}"
```
KennethNielsen, line 351: If you want to expand with an error message here, I think it should be more specific, like mention that it is the unit which doesn't match

tests/functional/test_eclab_parser.py
```diff
@@ -0,0 +1,406 @@
...
+    The method is testing whether the values of the parsed data
+    in the columns are the same.
+
+    """
+    measurement = measurements_with_data[0]
+    column_data = measurements_with_data[1]
+
+    # test first and last values in columns
+    for column_name, unit_values in column_data.items():
+        # FIXME test fails for this column in these files, because the cycle number
```
KennethNielsen, line 367: Good

KennethNielsen (changes requested):
Looks great. I really glad we will get this in.

There are a few minor requests for changes and then I think it is good to go.

tests/functional/test_eclab_parser.py
```diff
@@ -0,0 +1,406 @@
...
+        "Efficiency/%": ("%", (0.0000000e000, 0.0000000e000)),
+        "cycle number": ("", (0.000000000000000e000, 0.000000000000000e000)),
+        "P/W": ("W", (2.2328596e-008, 2.7340908e-008)),
+        "Ewe-Ece/V": ("V", (-2.5963283e-001, -2.2592688e-001)),
+        "loop_number": ("", (0, 1)),  # added by Ixdat
+    },
+}
+
+
+@fixture(scope="module", params=tuple(TEST_DATA.items()))
```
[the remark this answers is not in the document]
matenestor: Thanks, also new for me.

tests/functional/test_eclab_parser.py
matenestor: Well, the pytest is telling us that it is the test for units that failed, but it does not say which column has the different unit value. So I wanted to make it more clear which column to look at.
I will make the assert messages more specific.

KennethNielsen:
@matenestor I forgot one thing. Could you add a teeny tiny section to this doc: https://github.com/ixdat/ixdat/blob/main/docs/source/developing.rst in a testing section, describing the use of submodules for extended tests and the need to initialize and update submodules if one wants to run those extended tests?

docs/source/developing.rst
```diff
@@ -101,8 +101,36 @@ We use tools to make sure that our code is both functional and pretty. This make
...
+the tests should be **marked** as ``external`` with the pytest mark decorator.
+It is done like this, so other developers do not have to download your data
+through submodules, when they do not wish to test their new feature with your
+tests and data.
+
+If you want to run the complete test suite that includes external tests from
+all contributing developers, you need to initialize necessary submodules::
+
+    git submodule update --init --recursive
+
```
ScottSoren, line 121: Here we could also mention:
"""
To download a specific submodule (here the ixdat-large-test-files submodule), the command is
`git submodule update --init submodules/ixdat-large-test-files`
"""

docs/source/developing.rst
```diff
@@ -101,8 +101,36 @@ We use tools to make sure that our code is both functional and pretty. This make
...
+And then run command for testing with ``--external`` option::
```
ScottSoren, line 122: Is there a workflow whereby I have my test data in a submodule and run the tests on my new feature but don't download other testdata or run other external tests?

I think no, because there's only the `--external` tag.
Should the relevant tests not be marked with `--external` until a feature branch is merged? And when should the git submodule be created? What is the intended workflow there?

ScottSoren (approved):
Hey @matenestor , this looks great! I've tried it out from your fork and it works as I think intended, with the data files not downloading on the initial clone, but only when I've asked git for the submodule.
The tests look good! I'll take a look later at what's happening with `cycle_number`.

I've added a couple questions about workflow, but I think this is good to merge.

matenestor:
Let's wait with the merge yet, because I will probably need to change some test data in the tests, since you made fixes in 93b3c6b.
And I might have one more test coming. It is for joined exported measurements to csv file, though.

docs/source/developing.rst
matenestor: Good question, yes there is. Executing only some external tests or a test with data from just some submodules can be achieved with pytest `-k` option -> [keyword expression](https://docs.pytest.org/en/7.1.x/how-to/usage.html#specifying-which-tests-to-run).

E.g. to run the tests from this PR only, run this: `pytest tests --external -k eclab`. In case there is more Python modules with "eclab" in their name, or when one wants to be extra specific, run `pytest tests --external -k test_eclab_parser`. Specifying an expression with `-k` works even for classes, methods and functions.

matenestor:
@ScottSoren I run the test with the new fixes and updated the code accordingly. I had to merge the `ecms-debugging` branch. Now it is ready to merge. (Unfortunately, no new tests were added.)

PR #100, opened 2022-09-28
New Zilien reader

matenestor:
The PR introduces a new Zilien reader for the SpectroInlets' new Zilien dataset file format version. The new dataset version is able to integrate the Biologic EC-lab dataset during a measurement. Such dataset contains a new series with header `EC-lab` with two meta columns `experiment_number` and `technique_number` that the new reader is using during the process of creating Ixdat series objects. The objects created from the Zilien dataset match exactly the objects created from the Biologic mpt files, like when read one by one.

All the test files from `test_data/Zilien version 1` had been moved to the _ixdat-large-test-files_ repository. They are now available via git submodule.

This PR depends on the 3rd PR in the _ixdat-large-test-files_ repository, if you want to use `--external` files.


src/ixdat/readers/zilien.py
```diff
@@ -162,150 +177,279 @@ def read(self, path_to_file, cls=ECMSMeasurement, name=None, **kwargs):
...
+
+    @staticmethod
+    def _get_series_splits(series_headers):
+        """Series names and their index pairs in the whole read dataset.
+
+        Args:
+            series_headers (list): Series name headers (even empty).
+
+        Returns:
+            list, list: List of tuples with index pairs
```
KennethNielsen, line 358: er list, list or list of tuples?

src/ixdat/readers/zilien.py
```diff
@@ -162,150 +177,279 @@ def read(self, path_to_file, cls=ECMSMeasurement, name=None, **kwargs):
...
+        # Create Ixdat series
+        column_series = []
+        for begin, end in splits:
+            column_series += self._create_series_objects(
+                column_headers, names_and_units, data[begin:end, :]
+            )
+
+        return column_series
+
+    def _create_series_objects(self, column_headers, names_and_units, data):
```
KennethNielsen, line 309: I could maybe use some argument descriptions, or a different name for data, indicating that it is already row cut

src/ixdat/readers/zilien.py
```diff
@@ -162,150 +177,279 @@ def read(self, path_to_file, cls=ECMSMeasurement, name=None, **kwargs):
...
+
+        for column_number, column_header in enumerate(column_headers):
+            # Skip meta columns in the EC-lab dataset
+            if column_header in ("experiment_number", "technique_number"):
+                continue
+
+            column_data = data[:, column_number]
+
+            # Skip holes in the EC-lab dataset
+            if all(np.isnan(column_data)):
```
KennethNielsen, line 323: This can be pushed all down to numpy with `all` on the array: https://stackoverflow.com/a/60677862

src/ixdat/readers/zilien.py
```diff
@@ -162,150 +177,279 @@ def read(self, path_to_file, cls=ECMSMeasurement, name=None, **kwargs):
...
-                column_series = TimeSeries(**series_kwargs, tstamp=timestamp)
-                last_time_series = column_series
+
+            # Create the series
+            if column_header in ("Time [s]", "time/s"):
+                series_object = TimeSeries(**series_kwargs, tstamp=self._timestamp)
+                time_series = series_object
             else:
-                column_series = ValueSeries(**series_kwargs, tseries=last_time_series)
+                assert (
```
KennethNielsen, line 339: As above

src/ixdat/readers/zilien.py
```diff
@@ -162,150 +177,279 @@ def read(self, path_to_file, cls=ECMSMeasurement, name=None, **kwargs):
...
+        series_cut_indices = [
+            (begin, end)
+            for begin, end in zip_longest(series_cut_indices, series_cut_indices[1:])
+        ]
+
+        return series_cut_indices, nonempty_headers
+
+    @staticmethod
+    def _get_biologic_splits(technique_numbers):
+        """Index pairs of row splits in the biologic dataset columns.
```
KennethNielsen, line 381: Please update this on the other docstrings where it ids not the casem to follow pep257 https://peps.python.org/pep-0257/ so describe what it does, not what it returns

KennethNielsen: > Notes:
> 
>     Triple quotes are used even though the string fits on one line. This makes it easy to later expand it.
>     The closing quotes are on the same line as the opening quotes. This looks better for one-liners.
>     There’s no blank line either before or after the docstring.
>     The docstring is a phrase ending in a period. It prescribes the function or method’s effect as a command (“Do this”, “Return that”), not as a description; e.g. don’t write “Returns the pathname …”.
>     The one-line docstring should NOT be a “signature” reiterating the function/method parameters (which can be obtained by introspection). Don’t do:

tests/functional/test_zilien_reader.py
```diff
@@ -119,7 +130,52 @@
     ),
 }
 
+ZILIEN_COLUMNS_AMOUNT = 56
+
+
+# fmt: off
+@fixture(scope="class", params=(
```
KennethNielsen, line 137: So a parametrized fixture? Maybe that deserves a brief comment, have never seen that before, but cool

tests/functional/test_zilien_reader.py
```diff
@@ -168,3 +225,71 @@ def test_read(cls_or_technique, expected_series):
...
+
+
+class TestZilienIntegrated:
+    """Tests for reading the Zilien files with integrated Biologic dataset."""
+
+    @pytest.mark.external
+    def test_series_match(self, datasets):
+        """Test presence of all the series from the mpt files in the tsv file."""
+
+        # coming from parametrized fixture
```
KennethNielsen, line 237: ah, here they are

tests/unit/test_zilien_reader.py
```diff
@@ -0,0 +1,385 @@
...
+def patch_biologic_dataset_part(column_headers_cut, data_cut):
+    """Patch function for a biologic part call inside form series method."""
+    return [
+        (column_header, data_cut[0, i])
+        for i, column_header in enumerate(column_headers_cut)
+    ]
+
+
+def patch_zilien_dataset_part(series_header, column_headers_cut, data_cut):
+    """Patch function for a zilien part call inside form series method."""
```
KennethNielsen, line 24: Again as per PEP257

KennethNielsen (changes requested):
This looks really great. Nice and clean code for the parsing an good tests. There are a few requests for changes, but nothing major.

src/ixdat/config.py
```diff
@@ -31,7 +31,7 @@ class _Config:
     def __init__(self):
         self.standard_metadata_suffix = ".ix"
         self.standard_data_suffix = ".ix.npy"
-        self.standard_data_directory = Path.home() / "ixdat"
+        self.standard_data_directory = Path.home() / ".ixdat"
```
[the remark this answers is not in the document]
matenestor: I discussed this with @ScottSoren and he agreed to make it hidden. The data is still easily accessible and this directory is not used by the users directly, in my opinion. It is the Ixdat using the data inside and I think it is rude when you download a package and it starts polluting your home directory.

src/ixdat/readers/zilien.py
```diff
@@ -162,150 +177,279 @@ def read(self, path_to_file, cls=ECMSMeasurement, name=None, **kwargs):
...
-            timestamp (float): The timestamp of the measurement
-            series_headers (List[str]): List of series headers, slots with empty strings
-                means the same as the last non-empty one
-            column_headers (List[str]): List of column headers
+        # The first 4 lines always include the file version, number of header lines,
+        # number of data header lines and data start line in this order.
+        # Backwards compatibility is ensured, because the one extra line read will be
+        # just added to the 'metadata' dict and ignored later.
+        metadata = {}
+        fixed_metadata_lines_amount = 4
```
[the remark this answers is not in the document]
matenestor: Yeah, I was thinking that if there is not metadata (and data), then the dataset is useless anyway and will not work.

src/ixdat/readers/zilien.py
```diff
@@ -162,150 +177,279 @@ def read(self, path_to_file, cls=ECMSMeasurement, name=None, **kwargs):
...
+        series_cut_indices, nonempty_headers = self._get_series_splits(
+            self._series_headers
+        )
 
-        # Iterate over numbered series and column headers
-        for column_number, (series_header, column_header) in enumerate(
-            zip(series_headers, column_headers)
-        ):
-            last_series_header = series_header or last_series_header
+        assert len(nonempty_headers) == len(
```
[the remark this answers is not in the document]
matenestor: I think I forgot to change it.. will do

src/ixdat/readers/zilien.py
matenestor: Both are lists of tuples. Do you think I should write out all types the return value consists of?

src/ixdat/readers/zilien.py
matenestor: Nice. I seriously need to learn all the numpy magic.

tests/functional/test_zilien_reader.py
matenestor: I did it also in my previous PR #86 here (https://github.com/ixdat/ixdat/pull/86/files#r908083418).

src/ixdat/config.py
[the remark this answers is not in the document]
KennethNielsen: Well if you agreed I won't complain more.

> It is the Ixdat using the data inside and I think it is rude when you download a package and it starts polluting your home directory.

Sure, except I would argue, more so, if it dumps **significant** amounts of data in a hidden folder

src/ixdat/readers/zilien.py
KennethNielsen: Ah, my bad, I read read it wrong. No, don't write the more out, this is fine.

tests/functional/test_zilien_reader.py
KennethNielsen: Yeah, it is great. I think you also commented on it below, so probably just ignore this

src/ixdat/readers/zilien.py
matenestor: I renamed the variable and wrote more detailed doc string.

src/ixdat/config.py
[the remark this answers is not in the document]
matenestor: There could be a warning in the main readme or docs. But I think the dumping is inevitable in both ways, when the user needs the data.

tests/functional/test_zilien_reader.py
ScottSoren: This also threw me off. I could use a comment though on what is in this particular fixture.
Maybe mention that it is `(path_to_tsv, paths_to_mpts, time_offsets)` and why that is a useful thing to use in tests.

tests/functional/test_zilien_reader.py
```diff
@@ -168,3 +225,71 @@ def test_read(cls_or_technique, expected_series):
...
+        tsv, mpts, _ = datasets
+
+        # if one name from a mpt file would match a name from a different mpt
+        # in the integrated tsv dataset, then the test for series amount will
+        # fail, so it is not necessary to check for the exact belonging
+        for mpt in mpts:
+            for mpt_name in mpt.series_names:
+                if mpt_name == "time/s":
+                    assert f"Biologic {mpt_name}" in tsv.series_names
```
ScottSoren, line 246: I don't get this. How does this end up in `series_names`? Why put it there?

tests/functional/test_zilien_reader.py
```diff
@@ -3,20 +3,31 @@
...
     MSMeasurement,
     ECMeasurement,
 )
 
-DATA_DIR = Path(__file__).parent.parent.parent / "test_data"
 
-PATH_TO_DATAFILE = DATA_DIR / "Zilien version 1" / "2022-04-06 16_17_23 full set.tsv"
+DIR_VERSION_1 = (
+    Path(__file__).parent.parent.parent
+    / "submodules/ixdat-large-test-files/zilien_version_1"
```
ScottSoren, line 23: I don't see these in the ixdat-large-test-files repository. Please make that PR.
Looking at it here now: https://github.com/matenestor/ixdat-large-test-files

tests/unit/test_zilien_reader.py
```diff
@@ -0,0 +1,385 @@
...
+        ]
+        data = np.array(
+            [
+                [0.1, 1, 3, "NaN"],
+                [0.2, 1, 4, "NaN"],
+            ],
+            dtype=float,
+        )
+
+        series = reader._create_series_objects(column_headers, names_and_units, data)
```
ScottSoren, line 58: I don't see the logic of this function - why it needs four column headers and four tuples of names_and_units to make two series.

src/ixdat/readers/zilien.py
```diff
@@ -162,150 +177,314 @@ def read(self, path_to_file, cls=ECMSMeasurement, name=None, **kwargs):
...
+        # Create Ixdat series
+        column_series = []
+        for begin, end in splits:
+            column_series += self._create_series_objects(
+                column_headers, names_and_units, data_columns_cut[begin:end, :]
+            )
+
+        return column_series
+
+    def _create_series_objects(self, column_headers, names_and_units, data_rows_cut):
```
ScottSoren, line 332: When I saw this tested in the Zilien reader unit test I realized I do not understand what it is doing. Can you mention in the docstring here that it skips `experiment_number` and `technique_number`, and why?

ScottSoren (changes requested):
Hej Jakub,
Looks like a nice development in Zilien's tsv file and it's reader.
First of all, I was disapointed to see that my main critique of the tsv file remains - that it doesn't print the timestamp in the file. It annoyingly still relies on the timestamp being in the title. This is annoying to the user, because it means we can't rename the files without breaking ixdat's ability to read it. If reforming the .tsv file, this is a very low-hanging fruit and I can't see any reason not to put the tstamp in the file header.

Apart from that, it looks great as far as I can tell. Most of my comments have to do with not quite understanding the implementation. I didn't have time to play with it myself, so was only relying on the code and docstrings.
First of all, the NEXT_CHANGES.rst should explain a bit more. API changes should be accompanied by a code example. You state that if one reades a new Zilien file and gets the biologic part, that it is an identical measurement to if one just reads and appends the biologic files. However, it is not obvious for me what would be the code to do this? The only thing that resembled that I noticed was `ZilienTSVReader._biologic_dataset_part()` but that can't be right because it returns a list of `DataSeries` rather than a measurement, and it is a hidden method. 
Should there actually be a `ZilienMeasurement(ECMSMeasurement)` class that has user-facing methods for this kind of thing? And also reads the spectra? Maybe some of that is out of scope, but I could use something helping me to understand what and why and where it's going. I'm also a bit confused about the inner workings - temporary stand-in DataSeries called "Biologic XX" seem to be made and then filled out with the actual data later? This is referred to a couple places, but I couldn't understand how it works with the time available, which might indicate some docstrings could be kinder to the code reviewer.

src/ixdat/readers/zilien.py
```diff
@@ -30,6 +31,8 @@
     ECMeasurement: ZILIEN_EC_ALIASES,
 }
 
+BIOLOGIC_SERIES_NAME = "EC-lab"
+
 # TODO: When, in the future, Zilien files include the whole EC dataset, remove the
```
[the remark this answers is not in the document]
KennethNielsen: Yes. I actually told Jakub that the documentation chsanges could wait. Because I wanted to land the code changes soon, so that is on me.

src/ixdat/readers/zilien.py
```diff
@@ -135,25 +148,27 @@ def read(self, path_to_file, cls=ECMSMeasurement, name=None, **kwargs):
 
         # Extract timestamp from filename on form: 2021-04-20 11_16_18 Measurement name
         file_stem = self._path_to_file.stem  # Part of filename before the extension
-        timestamp = timestamp_string_to_tstamp(
+        self._timestamp = timestamp_string_to_tstamp(
```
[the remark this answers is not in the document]
KennethNielsen: Good point. I'm sorry I missed that. We **do** actually add the timestamp as a metadata item now. It was always supposed to be there and it was a bug that it was missed in the first place. Point is that now, if the timestamp is in metadata we should use that, if not, then we fall back to using the one from the file name.

tests/functional/test_zilien_reader.py
matenestor: The PR is already opened here https://github.com/ixdat/ixdat-large-test-files/pull/3.

tests/unit/test_zilien_reader.py
matenestor: Beside testing the series creation, it is also testing that `experiment_number` and `technique_number` columns, and columns with only "NaN" values are skipped.

tests/functional/test_zilien_reader.py
matenestor: The `Biologic time/s` is a column with timestamps from the Biologic measurement. So it needs to be created as a `TimeSeries` that is used as a `tseries` attribute for the following `ValueSeries`.

tests/functional/test_zilien_reader.py
matenestor: I described it more in the doc string and changed the variable name, where the parameters are unpacked, from `mpt_offsets` to `mpt_time_offsets`.

src/ixdat/readers/zilien.py
matenestor: Done. It skips them, because they are not in the MPT files and Ixdat does not create them as well (doesn't have from what). We want to create the same series, not to add extra unnecessary series.

src/ixdat/readers/zilien.py
```diff
@@ -162,150 +177,314 @@ def read(self, path_to_file, cls=ECMSMeasurement, name=None, **kwargs):
...
+            # Is the column a "setpoint" or "value" type
             setpoint_or_value = None
             for option in ("setpoint", "value"):
                 if series_header.endswith(option):
                     setpoint_or_value = option
 
+            mass = to_mass(series_header)
             if setpoint_or_value:
                 # In that case, the column header is something like "Flow [ml/min]" where
                 # "Flow" is unnecessary, because that is apparent from the unit
```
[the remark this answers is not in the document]
matenestor: Done, and the `elif` branch also.

src/ixdat/readers/zilien.py
```diff
@@ -162,150 +177,314 @@ def read(self, path_to_file, cls=ECMSMeasurement, name=None, **kwargs):
...
+        ]
+
+        return series_cut_indices, nonempty_headers
+
+    @staticmethod
+    def _get_biologic_splits(technique_numbers):
+        """Create index pairs of row splits in the biologic dataset columns.
+
+        Args:
+            technique_numbers (np.array): Numbers of techniques during
```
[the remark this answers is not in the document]
matenestor: I can't use that, because no `Measurement` exists yet. It might return the same numbers, but it is just a different solution. Also, I don't want to create series in the background and confuse a user with not consecutive ids.

src/ixdat/readers/zilien.py
```diff
@@ -87,6 +94,12 @@ def __init__(self):
         self._cls = None
         self._measurement = None
 
+        self._timestamp = None
+        self._metadata = None
```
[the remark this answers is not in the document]
matenestor: It is explained a bit now. And an example of the new Zilien file is here in the new PR https://github.com/ixdat/ixdat-large-test-files/pull/3/files#diff-fd13766ea72f1171d0eb4624236e2d9a984b46a6654c4178a8e3b051ef0f8890.

src/ixdat/readers/zilien.py
```diff
@@ -162,150 +177,314 @@ def read(self, path_to_file, cls=ECMSMeasurement, name=None, **kwargs):
...
+
+        count = self._metadata[f"{BIOLOGIC_SERIES_NAME}_{BIOLOGIC_SERIES_NAME}_count"]
+        names_and_units = [
+            self._form_names_and_unit(BIOLOGIC_SERIES_NAME, column_header)
+            for column_header in column_headers
+        ]
+
+        # Split rows according to techniques used according to experiments
+        # Combine the experiment numbers and the technique numbers
+        # in order to create unique identifiers for a successful split
```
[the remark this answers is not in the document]
matenestor: Yes, can't use it, because it is in the process of building the measurement.

src/ixdat/readers/zilien.py
```diff
@@ -135,25 +148,27 @@ def read(self, path_to_file, cls=ECMSMeasurement, name=None, **kwargs):
...
-            metadata, series_headers, column_headers = self._read_metadata(file_handle)
+            (
+                self._metadata,
+                self._series_headers,
+                self._column_headers,
+            ) = self._read_metadata(file_handle)
             file_position = file_handle.tell()
 
         # Read raw data
         with open(self._path_to_file, "rb") as file_handle:
```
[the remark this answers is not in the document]
matenestor: The PR is there https://github.com/ixdat/ixdat-large-test-files/pull/3.

I don't know exactly why it needs to be read as binary, because it is not my code. Probably, the `np.genfromtxt()` method prefers to read it as a binary instead of text, because there is a `file_handle.seek()` used. Maybe @KennethNielsen can explain more.

test_data/Zilien version 1/full set mass scans/mass scan started at measurement time 0000026.tsv
```diff
@@ -1,261 +0,0 @@
-num_header_lines	Number of header lines		int	8
```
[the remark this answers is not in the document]
matenestor: There is no class like `SpectroECMSMeasurement`, so I am confused (not my code). There is a `"zilien_spec": ZilienSpectrumReader`, maybe it is doing something. @KennethNielsen ?

src/ixdat/readers/zilien.py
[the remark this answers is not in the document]
matenestor: Now, it will check the unix timestamp in the file metadata first. Only when it is not there it will try to parse it from the filename.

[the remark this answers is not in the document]
matenestor: Although the difference between these two is about 3 seconds. I am finding out why in our codebase now,

NEXT_CHANGES.rst
```diff
@@ -49,3 +49,14 @@ The workings of ``MSMeasurement.grab_flux`` are changed if ``plugins.USE_QUANT``
...
+New Zilien reader
+=================
+
+A new Zilien reader for the SpectroInlets' new Zilien dataset file format
+version. The new dataset version is able to integrate the Biologic EC-lab
+dataset during a measurement. Such dataset contains a new series with header
+EC-lab with two meta columns experiment_number and technique_number that the
+new reader is using during the process of creating Ixdat series objects. The
+objects created from the Zilien dataset match exactly the objects created from
+the Biologic mpt files, like when read one by one.
```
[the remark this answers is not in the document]
matenestor: Usage is the same like it was before. There will be just more series, when the TSV file includes also Biologic data.

When you want to get only the Biologic part, you need to create a `Measurement` from the TSV file, then grab by the `Biologic time/s` series and according to that take all related series that have such `tseries`.

Maybe there is a better way in the Ixdat. At this point, it is not possible to read the file and get **only** the Biologic part. It will always include our Zilien data. I don't see a point of reading a Zilien dataset, when you want only the Biologic data. You can simply `read_set()` directory with the MPT files.

matenestor:
> You state that if one reades a new Zilien file and gets the biologic part, that it is an identical measurement to if one just reads and appends the biologic files. However, it is not obvious for me what would be the code to do this? The only thing that resembled that I noticed was ZilienTSVReader._biologic_dataset_part() but that can't be right because it returns a list of DataSeries rather than a measurement, and it is a hidden method.

 The `ZilienTSVReader._biologic_dataset_part()` is exactly what creates the **series** from the TSV file data, that are later accessible in the `Measurement` **object**. You don't get one `Measurement` for the Biologic part and another for the Zilien part. One Zilien TSV file produces one `Measurement` with **series** from both the Zilien and the Biologic datasets.

> Should there actually be a ZilienMeasurement(ECMSMeasurement) class

The `Measurement` classes are Ixdat's thing. I will not attack its main architecture. The default class is `ECMSMeasurement` which fits, because we always have potentiostat data inside. Furthermore, when you have the Biologic dataset included, it needs to read either as `ECMSMeasurement` or just `ECMeasurement`. Otherwise it will not get included, logically.

> temporary stand-in DataSeries called "Biologic XX" 

There is only one header name with a prefix "Biologic" and that is `Biologic time/s`. It gets prefixed in order to be different from the `Potential time/s` and other massspec `Time [s]` columns.

src/ixdat/readers/zilien.py
```diff
@@ -167,6 +165,20 @@ def read(self, path_to_file, cls=ECMSMeasurement, name=None, **kwargs):
             file_handle.seek(file_position)
             self._data = np.genfromtxt(file_handle, delimiter="\t")
 
+        # Part of filename before the extension
+        file_stem = self._path_to_file.stem
+
+        if "start_time_unix" in self._metadata:
```
ScottSoren, line 171: Awesome!

src/ixdat/readers/zilien.py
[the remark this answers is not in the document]
ScottSoren: As in, it's 3 s slower to read the tstamp from inside the file than from the title??

src/ixdat/readers/zilien.py
```diff
@@ -2,6 +2,7 @@
```
ScottSoren, line 2: Could you add a bit to the module docstring here about the quirks of Zilien files? 
For instance, can you define here the terminology `_series_headers` and `_column_headers` which is non-obvious, and what confused me last time. I think Zilien .tsv is kind of unique among data files I've seen in having both of these, aka a multi-line data column definition, so best to define this terminology for the reader of the code. 
Add (with changes as you see fit):
"""
Zilien tsv files are somewhat unique in having two lines to define each of the data columns. The first one is referred to as "**series header**" and explains what the data describes, and the second one is called "**column header**" and specifies the speficic column. Typically a series header will specify the measuring device (e.g. "iongauge value") or MS channel (e.g. "C0M2") and will apply for two or more column headers specify the quantity where the first is time ("Time [s]") and the subsequent are the corresponding value(s) ("Pressure [mbar]" or "M2-H2 [A]"). 
Newer zilien files can also include all the data from a coupled Biologic potentiostat. These are grouped under the **series header** "EC-lab".
"""

src/ixdat/readers/zilien.py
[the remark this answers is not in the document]
ScottSoren: no problem. Initiating documentation overhaul is an overdo thing on me ever since workshops in May.

src/ixdat/readers/zilien.py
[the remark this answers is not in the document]
ScottSoren: Thanks for pointing me to the example data. @KennethNielsen , is there a simple explanation of why this is read as binary that we could put into a comment here?

src/ixdat/readers/zilien.py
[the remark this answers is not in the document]
ScottSoren: I went ahead and merged ixdat-large-test-files/pulls/3. However git submodule update doesn't reflect that change. Looking at it on github.com, it the submodule to be pinned to a specific commit. So I guess that should be changed in this PR? (For now I manually cloned the ixdat-large-test-files package into submodules/..)

src/ixdat/readers/zilien.py
```diff
@@ -87,6 +94,17 @@ def __init__(self):
         self._cls = None
         self._measurement = None
 
+        # start time of the Zilien measurement
+        self._timestamp = None
+        # a dictionary with metadata general information about the Zilien measurement
+        self._metadata = None
+        # a list with the Zilien TSV series headers
+        self._series_headers = None
+        # a list with the Zilien TSV columns headers
```
ScottSoren, line 103: Sorry to be annoying, but can you add just a bit more here? The names `_series_headers` and `_column_headers` really threw me off. An example in the comment would help: 
```
        # a list with the Zilien TSV series headers, such as "Ionguage value" (see module docstring)
        self._series_headers = None
        # a list with the Zilien TSV columns headers, such as "Time [s]" and "Pressure [mbar]"
        self._column_headers = None
        # a numpy array of the parsed Zilien data, which is a big rectangle with Nan filling
        self._data = None
```

src/ixdat/readers/zilien.py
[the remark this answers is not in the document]
ScottSoren: good point.

src/ixdat/readers/zilien.py
```diff
@@ -162,150 +189,321 @@ def read(self, path_to_file, cls=ECMSMeasurement, name=None, **kwargs):
...
+        series = []
 
-        # Iterate over numbered series and column headers
-        for column_number, (series_header, column_header) in enumerate(
-            zip(series_headers, column_headers)
-        ):
-            last_series_header = series_header or last_series_header
+        # Get non-empty series headers and their indices
+        # in order to process whole series chunks
+        series_cut_indices, nonempty_headers = self._get_series_splits(
```
ScottSoren, line 238: Would you mind refactoring this to `series_split_indeces`? As an unwritten (as of now) convention, "cut" in ixdat so far always refers to selecting in time (and "snip" refers to selecting in a spectrum's independent variable like wavelength or m/z). So that was confusing me here - I conldn't see how `series_cut_indices` was cutting anything with respect to time.

ScottSoren (changes requested):
Just looked at this PR myself and played with a bit using pycharm's debugger. It looks good, and I can follow what's going on! 

I've made a couple tiny suggestions, but otherwise it's good to merge by me! The main thing that had been confusing me was that I didn't know what "series_headers" and "column_headers" referred to until digging deep. So I've suggested helping the reader of the code with that.

I went ahead and merged https://github.com/ixdat/ixdat-large-test-files/pull/3 already. However git submodule update doesn't reflect that change. Looking at it on github.com, it the submodule to be pinned to a specific commit. So I guess that should be changed here? (For now I manually cloned the ixdat-large-test-files package into submodules/..)

src/ixdat/readers/zilien.py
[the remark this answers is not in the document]
matenestor: No no, it is due to a synchronization of the file creation and a core initialization reasons. But now, the Zilien files will _always_ include the (correct) timestamp and it will _always_ be read from the file. So I am up for leaving the (wrong) timestamp in the filename purely for informative and sorting reasons. The time difference between the two timestamps depends on how fast a user's computer is.
Unfortunately it is not easily possible to put the correct timestamp into the filename.

src/ixdat/readers/zilien.py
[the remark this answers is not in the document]
matenestor: Yes, I needed the PR in the `ixdat-large-test-files` repo to be merged in order to update the submodule. Now, there is a new commit here that will update submodules to the recent version.

src/ixdat/readers/zilien.py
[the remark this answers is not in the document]
ScottSoren: Great!
Understood - that's the time it took Zilien to make the .tsv from the data in the tmp folder, then, I guess?

test_data/Zilien version 1/full set mass scans/mass scan started at measurement time 0000026.tsv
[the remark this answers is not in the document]
ScottSoren: `SpectroECMeasurement` is defined here:  https://github.com/ixdat/ixdat/blob/a566942aa72b82beebba331a765423538baf2a99/src/ixdat/techniques/spectroelectrochemistry.py#L12
For a tutorial using it, see here: https://github.com/ixdat/tutorials/blob/main/spectroelectrochemistry/spectroelectrochemistry_demo.ipynb (The technique my colleagues call SEC is covered by the inheriting `ECOpticalMeasurement`)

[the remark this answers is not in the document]
ScottSoren: But I think it's out of scope for this PR, so save for later :)

ScottSoren (approved):

src/ixdat/readers/zilien.py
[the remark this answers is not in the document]
KennethNielsen: No. Let me try and clear up a few things, because as it actually turns out the implications here are a little bigger than just an annoyance.

The timestamp in the filename is the time at which the folders are created at measurement start and is only used to make sure that every started measurement will get a unique folder/file-name, even with identical project and measurement names. This is different by a few seconds from the measurement timestamp, which is the timestamp designated as measurement time 0.0 (and what ixdat understands as the timestamp) after all the software components reports ready for measurement.

The annoying conclusion to this is that the timestamp in the filename **should not be used to synchronisation**, since that will give you a few seconds offset. The problem is, that since I missed this difference (I had forgotten that it was not the measurement timestamp in the filename), then the way that ixdat has done it so far has actually produced the wrong result and therefore measurements whose sync is of by a few seconds!!

For the new file version, all will be good, because we have taken care of the sync. But for the existing data sets I think what should happen, in order to make sure that the sync is good, is that when the separate MS part in a Zilien file gets combined with a EC-Lab measuement, that the Zilien measurement knows about, then it should adjust the timestamp of the EC-Measurement to that of the Zilien measurement + the offset in the metadata. This might be accomplished by some clever overloading the `__add__` but I will need to give it some more thought.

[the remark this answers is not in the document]
KennethNielsen: Even with this comment. I consider what I mentioned above a bug and will open an issue. I think it is all still correct for this PR so I will leave this comment resolved.

src/ixdat/readers/zilien.py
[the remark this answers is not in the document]
KennethNielsen: This is at the data part of the file, which means that we are passed the utf-8 encoded metadata. So I thought that I might as will cut the decode out of the picture and just hand it a binary file object. I thought it more efficient, but haven't tested. In any case, it doesn't really binary, as in tightly packed binary format, just that we don't need a special encoding.

KennethNielsen:
> I went ahead and merged https://github.com/ixdat/ixdat-large-test-files/pull/3 already. However git submodule update doesn't reflect that change. Looking at it on github.com, it the submodule to be pinned to a specific commit. So I guess that should be changed here? (For now I manually cloned the ixdat-large-test-files package into submodules/..)

You are correct. @matenestor now that the data-files PR is merged, you should add one more commit to this one that update the submodule to the latest version on Github so the new data is exposed.

src/ixdat/readers/zilien.py
KennethNielsen: Well, the first step is probably just to assume that if you are to do a common operation on an entire array, then most likely numpy will have a optimized version for it.

KennethNielsen (approved):

matenestor:
> You are correct. @matenestor now that the data-files PR is merged, you should add one more commit to this one that update the submodule to the latest version on Github so the new data is exposed.

Already done in f34cb1d. :)

ScottSoren (approved):
Looks good! Merging :)

PR #105, opened 2023-01-09
Fix tools

ScottSoren:
Hi @KennethNielsen ,

Our CI actions have been failing because of the py36 build in github. To get the green check again, I've here taken the stupid-simple approach and removed py36 (python 3.7, 3.8, and 3.9 are all still fine).

I also renamed the reader running scripts "demo" instead of "test" so that developers don't confuse them for software tests.


KennethNielsen (approved):
LGTM

PR #116, opened 2023-06-23
Quant integration

ScottSoren:
This includes the updates for using ixdat with the `spectro_inlets_quantification` (siq) package.

- See NEXT_CHANGES.rst for useage details
- See https://github.com/SpectroInlets/spectro_inlets_quantification/pull/1 for the changes that occured there
- demonstration is currently on unpublished data at https://github.com/ScottSoren/quant_integration

Note that molecule files were deleted here for inclusion (as yaml) in siq or in a data directory on the user's computer (by default "~/.ixdat/plugin_data/ms_quant"

In addition to all the siq stuff, it includes a handy `fix_WE_potential` function to repair data when Biologic exports CP data with 0's for Ewe/V.


src/ixdat/techniques/ms.py
```diff
@@ -92,10 +93,16 @@ def grab(
                 Defaults to False, but in grab_flux it defaults to True.
             include_endpoints (bool): Whether to ensure tspan[0] and tspan[-1] are in t
         """
+        if plugins.use_si_quant and item.startswith("n_dot_"):
+            return self.grab_flux(
+                item.lstrip("n_dot_"),
```
KennethNielsen, line 98: `strip` in Python **always** removes as many of the characters in the string you give as argument to `strip` it can find, **not** the as a substring. *Technically* I think you are fine, since no molecules start with any of the characters `["n", "d", "o", "t", "_"]`, but it is a bug since it doesn't work as intended and only works by accident, and should therefore be changed to `removeprefix`.

KennethNielsen (changes requested):
Looks great. There are only a few details on this.

I explicitly did **not** review the data files changes, since I assume that they were converted (and possibly checked) programmatically to be equivalent.

src/ixdat/techniques/ms.py
ScottSoren: thanks, I made a false assumption about stripping!. Changed to `removeprefix`.

ScottSoren:
Hi Kenneth, 

Thanks for the attentive typo & bug catches. I've fixed those, and also implemented some of the fall-out from Anna and my tutorial on Thursday. This involved:
- polishing `MSInlet.gas_flux_calibration_curve()` and copying it to `MSMeasurement`, where it returns a siqms `CalPoint` object. 
- deprecating `MSInlet`. The only MSInlet whos capillary equation we know is at Spectro Inlets, and therefore belongs in siqms, not ixdat.
- Modifying the `MSMeasurement.cut`, `MSMeasurement.multicut`, and `ECMSMeasurement.as_cv()` methods to copy over the quantifier. Implementing this in `multicut`  results in `select()` and indexing of `CyclicVoltammogram` objects to also carry over the quantifier.

The tutorial scripts (jupyter notebooks) and data that we worked with to motivate and test these changes is available here:
https://www.dropbox.com/scl/fo/59gglamqy1volvn3bmt8p/h?rlkey=irilsj4s2jjx7fjy8vhd70xsn&dl=0

src/ixdat/techniques/ms.py
```diff
@@ -348,7 +376,130 @@ def gas_flux_calibration(self, mol, mass, tspan, chip=None):
...
+        Return MSCalResult(, Axis): The result of the MS calibration (and calibration
+            curve axis if requested) based on flux calculation during selected time
+            periods.
+        TODO: automatically recognize the pressure from measurement (if available)
+        """
+        if not plugins.use_si_quant:
+            raise QuantificationError(
+                "`MSMeasurement.gas_flux_calibration` only works when using an "
+                "external MS quantification package "
+                "(`ixdat.options.use_si_quant = True`). "
```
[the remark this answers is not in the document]
AnnaWiniwarter: I also noticed the same further below in line 555 (but somehow cant add a comment there?!)

src/ixdat/techniques/ms.py
```diff
@@ -949,12 +1114,9 @@ def gas_flux_calibration_curve(
         mol,
         mass,
         tspan_list=None,
-        selector_list=None,
```
AnnaWiniwarter, line 952: Why are you removing this nice functionality of choosing EC selector periods rather than having to always pass a tspan? I would miss it dearly!

AnnaWiniwarter: Ah is it because it should be in EC-MS only?

AnnaWiniwarter: I'm also a bit confused as to why you have two definitions of the method `gas_flux_calibration_curve` in the same file?

AnnaWiniwarter (changes requested):
See comments. I am left a bit confused by the changes in ` techniques/ms`, so changes might not be necessary, but I'd like to understand before I approve. :) almost there!

src/ixdat/readers/biologic.py
```diff
@@ -34,6 +34,32 @@
...
+    Some Biologic potentiostats / EC-Lab versions sometimes export a column of zeros for
+    "<Ewe>/V" in the .mpt files in chronopotentiometry measurements. This function
+    replaces the series of zeros with the correct potential by adding the counter
+    electrode potential ("<Ece>/V") and cell potential ("Ewe-Ece/V").
+
+    Args:
+        measurement(ECMeasurement): The measurement with the column to be replaced
+    """
+    WE_series = measurement["<Ewe>/V"]
+    CE_data = measurement.grab_for_t("<Ece>/V", WE_series.t)
```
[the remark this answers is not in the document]
ScottSoren: This is only run manually. I've specified that now in the doc string, and added an error message in case it is used on a measurement for which the counter electrode potential was not recorded.

src/ixdat/techniques/ms.py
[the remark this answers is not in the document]
ScottSoren: Fixed both places! Nice catch :)  
(You coulden't comment on line 555 because it hadn't been changed wrt main)

NEXT_CHANGES.rst
```diff
@@ -15,6 +15,48 @@ links to relevant Issues, Discussions, and PR's on github with the following for
...
+
+Readers
+^^^^^^^
+- A function ``fix_WE_potential`` has been added to the ``biologic`` module. This
+  corrects the mistake in some .mpt files that are exported with zeros for "<Ewe>/V".
+
+Techniques
+^^^^^^^^^^
+- MS measurements now have access to the powerful ``spectro_inlets_quantification`` (siq)
+  package as a plugin. See `siq's documentation <https://github.com/SpectroInlets/spectro_inlets_quantification/blob/main/docs/source/theory/quant_ms.md>`_
```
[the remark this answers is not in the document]
ScottSoren: Hm... this confused me a bit, but it looks like this comment is outdated and I have fixed it since: https://github.com/ixdat/ixdat/pull/116/files
(make sure to select changes from "all commits" in top left, easy to miss!)

NEXT_CHANGES.rst
```diff
@@ -15,6 +15,48 @@ links to relevant Issues, Discussions, and PR's on github with the following for
...
+    return an object of ``CalPoint``.
+  - ``MSMeasurement.set_quantifier`` is used to attache a siq quantifier to the measurement.
+    Usage: ``ms.set_quantifier(calibration=my_siq_calibration)``
+  - ``MSMeasurement.grab_fluxes`` is a new method which uses the attached quantifier to
+    quantify the fluxes of all molecules covered therein. In general it solves a matrix
+    equation to deconvolute overlapping signals.
+  - ``MSMeasurement.grab_flux`` will, when siq is activated, run ``MSMeasurement.grab_fluxes``
+    and return the requested flux vector.
+
+  SIQ comes with data for a small batch of molecules. To supplement this of yaml-formatted
```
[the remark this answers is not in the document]
ScottSoren: same as above, I did implement it! github just isn't great at showing what was done _since_ a comment.

src/ixdat/techniques/ms.py
ScottSoren: For the first question: Yes! An `MSMeasurement` in general doesn't come with a selector...
For the second question: It's not elegant, but the first one (method of `MSMeasurement`) uses `spectro_inlets_quantification` and this one (method of `MSInlet`) uses ixdat's native MS calibration.

ScottSoren: I started to try refactoring in a way that moving the shared code to another function, but I haven't come up with a way of doing it that makes the module shorter, since there are so many arguments that need to get repeated anyway... happy if you give it a try! Will leave this comment unresolved.

ScottSoren: I just remembered, `MSInlet` is deprecated. I don't think it makes sense to have in ixdat, as the only inlet we have the capillary equation for is an SI chip, and that is covered in `spectro_inlets_quantification`. So next release won't have the repeated code :)

AnnaWiniwarter (approved):
All good now :D

ScottSoren:
Merging! :)

Issue #118, opened 2023-08-10
Wrong current co-plotted when highlighting current used for calibr. in ecms_calibration_curve


AnnaWiniwarter:
When co-plotting the current used in the calibration on a plot which has been calibrated to show current density by passing the axis to `ecms_calibration_curve`, the raw current will be co-plotted. of course, the raw current needs to be used for the calculation, but I think when highlighting in the plot, it would still make sense to highlight using the current density, just for looks of the plot. or do you think that's suggestive of the function actually using the current density and thereforte misleading? @ScottSoren 

Edit: if we dont want to allow it to highlight the current density, then I think we need to disallow it/give a warning if the current axis unit is not the same

ScottSoren:
This may be tricky, as the `ecms_calibration_curve` method plots that highlight on an axis passed to it. The information about what unit is plotted is not stored in the axis, so I'm not sure how the method could figure it out. Reading the axis label would be fragile, since the user could have changed it. Even implementation of `pint` in ixdat won't help - it would have to somehow be implemented in matplotlib! The best might just be to specify in the docstring that the axis passed to it should not have the current in [mA].

AnnaWiniwarter:
Ok, I didn't consider that the axis doesn't know what unit is plotted on it.

"The best might just be to specify in the docstring that the axis passed to it should not have the current in [mA]." - I'm not quite sure I understand what you mean by that. Do you mean we should be more specific that it won't work if calibrated data is plotted on the axis?

What about giving a possibility of passing to the method not only what axis to highlight, but also which unit this axis uses? i.e. not rely on it to be done automatically, but give the option to modify it?

ScottSoren:
acceptable fix for now, but see https://github.com/ixdat/ixdat/issues/130

Issue #94, opened 2022-07-22
Using "axes_measurement" to pass an axis to MSInlet.gas_flux_calibration_curve() broken


AnnaWiniwarter:
When passing an axis to MSInlet.gas_flux_calibration_curve() using axes_measurment, this should highlight the signal over the tspan selected to calculate the calibration. It doesn't work, however, when using the option of using a certain concentration in a carrier gas. Must have something to do with how the highlighted part is generated, seems it calculates the flux somehow?

AnnaWiniwarter:
I haven't tested it, but I would guess the same goes for single point gas calibrations.

KennethNielsen:
Any chance you can provide a working example and data that reproduces the bug?

ScottSoren:
There's an example here: https://github.com/ixdat/tutorials/blob/new_ecms_tutorials/ec_ms_quantification/EC-MS_ixdat_tutorial.ipynb

ScottSoren:
fixed.

PR #126, opened 2023-10-06
Fix ``axes_measurement`` issues in ``MSInlet.gas_calibration_curve()`` and ``ECMSMeasurement.ecms_calibration_curve``

AnnaWiniwarter:
This PR introduces small changes to the above methods, resolving issues #94 and #118. 
The approach is quite crude, especially for ECMSMeasurement.ecms_calibration_curve(), I'm sure there exists a better solution, so very open for suggestions.

I'm aware the branch currently doesn't pass tests. Need to install black etc on my new computer -  will do so after first round of review. :)

### DIFF
--- a/src/ixdat/techniques/ec_ms.py
+++ b/src/ixdat/techniques/ec_ms.py
@@ -132,6 +132,7 @@ class ECMSMeasurement(ECMeasurement, MSMeasurement):
         tspan_bg=None,
         ax="new",
         axes_measurement=None,
+        axes_measurement_J_name="raw_current",
         return_ax=False,
     ):
         """Fit mol's sensitivity at mass based on steady periods of EC production.
@@ -154,6 +155,11 @@ class ECMSMeasurement(ECMeasurement, MSMeasurement):
                 Defaults to a new axis.
             axes_measurement (list of Axes): The EC-MS plot axes to highlight the
                 ms_calibration on. Defaults to None. These axes are not returned.
+            axes_measurement_J_name (str): The J_name used in the axis passed 
+            to axes_measurement. Must be passed manually as the axis does not "know"
+            its J_name. Defaults to "raw_current". IMPORTANT: the method still uses
+            "raw_current" to calculate the sensitivity factor, this J_name is only
+            used for plotting.
             return_ax (bool): Whether to return the axis on which the calibration curve
                 is plotted together with the MSCalResult. Defaults to False.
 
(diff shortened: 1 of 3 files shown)


src/ixdat/techniques/ms.py
```diff
@@ -1189,12 +1194,16 @@ def gas_flux_calibration_curve(
             )
         else:
             cal_type = "gas_flux_calibration_curve"
+            mol_conc_ppm = 10**6
```
ScottSoren, line 1197: It seems this `mol_conc_ppm` never gets used!
Should it go into `mol_conc_ppm_list`?

src/ixdat/techniques/ms.py
```diff
@@ -1189,12 +1194,16 @@ def gas_flux_calibration_curve(
...
+            carrier_mol = mol
         for tspan, mol_conc_ppm, pressure in zip(tspan_list, mol_conc_ppm_list, p_list):
             t, S = measurement.grab_signal(mass, tspan=tspan, tspan_bg=tspan_bg)
             if axes_measurement:
-                axes_measurement.plot(t, S, color=STANDARD_COLORS[mass], linewidth=5)
-                mol_conc_ppm = 10**6
-                carrier_mol = mol
+                if axes_measurement_raw == True:
```
ScottSoren, line 1202: Don't use `==` when comparing to `True`. Use `is`, or if other stuff that evaluates as a boolean to `True` is okay, just use the thing itself. So here, `if axes_measurement_raw: `
flake8 enforces this.

src/ixdat/techniques/ec_ms.py
```diff
@@ -154,6 +155,11 @@ def ecms_calibration_curve(
                 Defaults to a new axis.
             axes_measurement (list of Axes): The EC-MS plot axes to highlight the
                 ms_calibration on. Defaults to None. These axes are not returned.
+            axes_measurement_J_name (str): The J_name used in the axis passed
```
ScottSoren, line 158: okay, but wouldn't it be more flexible to accept a J_unit_plot and then decide whether to use raw or normalized current based on whether it had a /cm^2 in it?
Maybe worth a revisit once ixdat has real units.

ScottSoren (changes requested):
Hi Anna! Looks good and it will be nice to resolve two issues! I've put a couple comments and questions. The one I think you should address before merge is the one about `mol_conc_ppm_list`.

ScottSoren:
I've now run black and checked with flake8, so it passes the checks..

src/ixdat/techniques/ms.py
ScottSoren: fixed.

src/ixdat/techniques/ms.py
AnnaWiniwarter: It was moved into the wrong place, should be fixed now.

ScottSoren:
Issue made for `J_unit_plot`, so we edit it when addressing units generally. Otherwise it looks great! Merging :)

ScottSoren (approved):

Issue #120, opened 2023-09-22
Executed EC-MS tutorial on documentation page misses code-text boxes


AnnaWiniwarter:
I was made aware of this issue recently - apparently all the code got lost when converting the executed tutorial to a .rst file. I believe this was already mentioned in #113 but I didn't take care of it then.

AnnaWiniwarter:
Turns out, the code is actually there in the file: docs/source/getting_started/tutorials/ecms_tutorial.rst
When I compile it locally, I also get the code pieces in the HTML file, but they are not on the compiled readthedocs page. Have you experienced something like this before @ScottSoren?

ScottSoren:
Seems something is broken with the `.. code:: ipython3` command.
I think we should just use `::` instead. That's what the documentation for ipython itself does: https://raw.githubusercontent.com/ipython/ipython/main/docs/source/interactive/tutorial.rst

I'll look into whether this is configurable in nbconvert: https://nbconvert.readthedocs.io/en/latest/usage.html#restructuredtext

ScottSoren:
I reproduce part problem: nbconvert puts the `code` directive instead of the (presumably) needed `code-block` directive. 

I suggest we try using nbsphinx instead: https://nbsphinx.readthedocs.io/en/0.9.3/index.html

ScottSoren:
Solved with #127 :)

Issue #115, opened 2023-05-17
Tutorials in documentation 


ScottSoren:
See https://github.com/ixdat/ixdat/pull/113#discussion_r1182547232

We need a way of keeping jupyter notebook tutorials version controlled, easily available for download, and visible in a readable form directly online - ideally directly documentation page. How to do this?

ScottSoren:
Solved with #127. 
Possible future improvement in #133

PR #127, opened 2023-10-12
Tutorials with nbsphinx

ScottSoren:
This is a request for comment pull request about how we do jupyter notebook tutorials. 

At present, we have [this tutorials repository](https://github.com/ixdat/tutorials) in which we can version control and code-review jupyter notebook tutorials. When doing workshops, I have also often gotten people to download tutorials from there. There are links in the README of that repository (and sometimes in the tutorial headers) to the data of that repository

At the same time, we have this documentation page which (after a few clicks) brings you to its collection of tutorials. The tutorials there are at present .rst exported by jupyter notebook, but something is broken about that export and sphinx meaning that the code cells have disappeared from the output. See https://github.com/ixdat/ixdat/issues/120

To solve this, I started looking at nbsphinx: https://nbsphinx.readthedocs.io/en/0.5.0/index.html
It is quite nice! It enables the jupyter notebook files to be directly here in the documentation source, to be compiled when building .html. I've suppressed output for now (in docs/conf.py) but the normal way for it to work is for it to run the jupyter notebooks and include the output in html. 

With some fix-up, this will enable us to just copy over the tutorials repository to the right place in the ixdat/docs folder and it shoud run! (solving https://github.com/ixdat/ixdat/issues/115) 

The organization of the tutorials has a lot of room to improve (see ideas from last May [here](https://www.dropbox.com/scl/fi/sijw8gjnazqgv2lpi7l47/22E19_WS1_Notes.txt?rlkey=mt27919dqxsgiviazge1f44r7&dl=0)) and some of the tutorials are broken. 

I was thinking we could solve some of those issues with PR's to this branch before merging it into main. But first let me know what you think of this use of nbsphinx!

You can download the html to preview here: https://www.dropbox.com/scl/fo/z9ea3pka9bjb31m334a3o/h?rlkey=90gp6qvdht2oo3rkb48buv2wx&dl=0


AnnaWiniwarter:
I think using nbsphinx sounds like a good idea. What I did previously was quite cumbersome (manually converting each jupyter notebook) and impractical for following up on changes to the notebooks, this sounds way better.

I had a look at the html you linked to above, and I'm not sure why you added a new section "jupyter notebooks" to the tutorials section? I agree, we need to fix/split/change some of the tutorials to make them more in line with the structure we agreed on previously, but the outline of this structure is already there in the current version of the documentation, and having both "tutorials" and "jupyter notebooks" is quite confusing I think.

Finally, it seems a bit dangerous to rely on manually copying updated jupyter notebooks from ixdat/tutorials to ixdat/ixdat/docs every time someone has made a modification. Is there really no smarter way of doing so?

ScottSoren:
@AnnaWiniwarter , thanks for the comments!

on workflow:
-------------
nbsphinx needs the notebooks to be in the documentation folder, so having it compile form notebooks in another repository doesn't work. Three options have come up for this:

1. notebooks only live in ixdat/docs . The tutorials repository becomes redundant and can be deleted. The main problem I see is that we need the tutorials not compiled (output cleared) for version control, but we need them compiled for the documentation site. We could actually have them un-compiled in ixdat/docs, but then sphinx will try to compile them for readthedocs, and if it crashes I think the whole ixdat.readthedocs.io site doesn't update. For it not to crash when sphinx compiles the notebooks on readthedocs' server, it would need to have all the data included. That's not the way we've done it so far. Notebooks can be set to not compile (That's what I did initially), but I think it's fair to say we want compiled notebooks in the end. 

2. Notebooks live in two places: a version-controlled version without output in the tutorials repository and a version with output in ixdat/docs. (At least the notebooks that require data too big for inclusion in the repository need to be upload with output, whereas we can choose to let sphinx compile self-contained notebooks.) The downside is having to copy them over. But I don't think this is bad. The whole repository can be copied over at once, making it a relatively risk-free operation. Selecting "keep most recent version" in case of conflicts ensures that jupyter notebooks that weren't edited don't need to be re-compiled. No re-organization should be needed since index.rst can define in what order / context they appear on the site. Copying over (and updating inxex.rst) becomes a way of "publishing" the present state of the notebooks repository. It could also be an invoke task, I think (@KennethNielsen)?

3. Tutorials can be a github sub-repository. @KennethNielsen mentioned this. I would be a bit challenged setting it up, and I can't figure out in my head right now whether it solves the with-output for documentation vs without-output for version control conflict mentioned above.

on page organization.
------------------------
I agree it is not implemented well. 

The intention is maybe more clear in index.rst: https://github.com/ixdat/ixdat/blob/tutorials_with_nbsphinx/docs/source/getting_started/tutorials/index.rst ... I think my idea was that the sphinx-compiled version of each notebook is linked to in text in the top part putting it in context, and then the dowloadable .ipynb links are listed at the bottom. The advantage of having a list is that it is easy to keep track of - the advantage of in-text links is that it is easy to read. But it seems I did almost the opposite - a table in the top and then the nbsphinx-converted notebooks in the bottom (and the .ipynb link kind of awkwardly hinging at the top). 

And one thing that surprised me a bit when I compiled it the first time is that the section headers in the markdown files of the .ipynb's get picked up by sphinx when it makes the table of contents! That seems to be what makes it look not at all like a list. It could be nice though if used right, since clicking the TOC links bring can bring you to specific parts of the notebooks!

Basically, I need more input on page organization. How would you have the tutorials page look? Just getting rid of the "jupyter notebooks" header and merging the top and bottom halfs that kind of repeat each other? And where should the links to the downloadable .ipynb go?

AnnaWiniwarter:
@ScottSoren thanks for your replies!
regarding **workflow**: I understand the challenge better now and option 2 sounds like the most sensible. If it can be an invoke task then I'd be less worried about missing out on updating the notebooks because of having to copy them from one place to another.
regarding **page organisation**: I dont really like having both the "sandwich" structure (i.e. the table) with links, and then a list of links to what seems to be the same things. I think the difference between the two is really not obvious from the titles "tutorial" and "jupyter notebook".
I think the embedded links in the table is the superior structure, and I also think that what you suggest makes sense (having separate direct links to the executed tutorials and the notebooks to download). What about keeping the previous "tutorials" page (renamed to "executed tutorials") and add a copy layout wise, which is called "jupyter downloads" (or tutorial downloads) where the links lead to the notebooks on the github page?

This doesnt solve what you mentioned about the automatic inclusion of headers in the TOC. I never figured out how to work well with that, so please let me know if you do. :)

KennethNielsen:
I just wanted to copy here even though the issue is closed.

With regards to workflow options 2 and 3. I think both can be done by adding some automation to an invoke task. Thoughts on characteristics as I see them.

**Subrepository**:

- A subrepository can "bound" to any folder in the repo, so it can be tied directly in inside the docs / source folder somewhere.
- A sub repository makes the link to another repository explicit, which means none of that "you have to have the other repository located in this specific location relative to this one for it to work.
- I **does** need to be manually updated when there are changes upstream, but it can be autodetected when these changes have appeared

As a final note here, I have sometime alluded to the fact that submodules in git have a bad reputation. As I understand from glancing, mostly the tech is sound, but the ergonomics of use is not the best. There is another git feature called subtree which is suggested to maybe be the best of submodules but with better usage. If we decide to go in this direction, I will do the research to figure it out properly. I will need this for other things as well.

**Seperate repo, copy over**

- This is a low tech solution that everyone will understand
- On the bad side the "link" exists only in documentation and invoke task definitions which is a little fragile.
- I'm unsure if this can be implemented on RTD, that is if we are allowed there to close an entirely different repo and copy stuff over. I think we can, but it will need to be checked.

ScottSoren:
@AnnaWiniwarter , I've fixed up the organisation on the Tutorials page of the documentation!

The html that it compiles now is here: https://www.dropbox.com/scl/fo/s5csxus8xq0s69rfx1aku/h?rlkey=97rxdcptjyw6gh5d3cr9jwucj&dl=0

Let me know if that looks better :) 

Lots of improvements to make on the tutorials themselves, and still some things to iron out with workflow, but it's getting there!

I think we should merge this ASAP and keep working on the rest after.

AnnaWiniwarter (approved):
Did not look at the code, but I think the doc page looks really really good now! The separation of the tutorials as a point on the main menu really makes sense considering the automatic generation of the sub-menues. So excited to see this go online! :)

AnnaWiniwarter:
Looking good still! I think it's ready to merge (and spread the message of a new and better doc page!) :D

ScottSoren:
Merged into main :) 
I'll set the default page on ixdat.readthedocs.io to be from ixdat's main again.

#133 will be left open to remind us we want a way to automate the workflow, but for now workflow is:
- work on jupyter notebooks in the tutorials repository (https://github.com/ixdat/tutorials) with output cleared
- when tutorials are finished, reviewed, and merged to the main branch of the tutorial repo, add them to the documentation:
- - make a branch or fork of the ixdat repository
- - copy the new/updated tutorials (or the whole tutorials repo, keeping only the most recent version of the files) into ixdat/docs/source/tutorials/tutorials_repo
- - execute the new/updated ipynb in the new location
- - make a pull request to the ixdat repository's main branch
- when the updated documentation is merged into the main branch, it will automatically compile on readthedocs. The new tutorial will have been converted to beautiful html.

Issue #93, opened 2022-07-20
Issue using cut when read in "empty" mpt data


KimDegnJensen:
**Fun little error, ixdat build-in functions are perfectly fine filling in an empty  .mpt experiment (txt file w. preamble but no actual data). The package will allow someting like this without any error**
data_dir_MS =  Path(r"experiment.tsv")
data_dir_EC_1 =  Path(r"experiment_mpt_file_1.mpt")
data_dir_EC_1 =  Path(r"experiment_mpt_file_1.mpt")
data_dir_EC_2 =  Path(r"experiment_mpt_file_2.mpt") #this is a "empty" mpt file which have been exported by error
data_dir_EC_3 =  Path(r"experiment_mpt_file_3.mpt") 
data_MS = ixdat.Measurement.read(data_dir_MS, technique="MS", reader="zilien")
data_EC_1 = ixdat.Measurement.read(data_dir_EC_1, reader="biologic")
data_EC_2 = ixdat.Measurement.read(data_dir_EC_2, reader="biologic")
data_EC_3 = ixdat.Measurement.read(data_dir_EC_3, reader="biologic")

full_ECMS_data=data_MS+data_EC_1+data_EC_2+data_EC_3

axes = full_ECMS_data.plot_measurement()
full_plot = axes[0].get_figure() 

**Data "looks" fine and is plottable, but using cut tool as such**
relevant_ECMS_data = ful_ECMS_data.cut(tspan=[tmin,tmax]) 

 **this promts an error looking like this**

  File "C:\...\anaconda3\lib\site-packages\ixdat\measurements.py", line 978, in tspan
    t_start = min(t_start, t[0]) if t_start else t[0]

IndexError: index 0 is out of bounds for axis 0 with size 0

**Other functions like grab etc. still work, so 1) maybe its only the cut function which have problem with the establishing a meaningfull t array 2) its hard for me to figure out whihc functions are affected 3) maybe the easiest way to solve this is stopping empty experiments to be readable e.g. by prompting a warning sayin "file X1 and X2" was empty so have been ignored**

ScottSoren:
Nice bug catch!
I would prefer to keep empty files readable, since this spares the user the headache of going through and deleting empty files when using a loop or `read_set` which reads a bunch. 

[...]

PR #131, opened 2023-10-31
solve #93

ScottSoren:
This fixes the issue that accessing `meas.tspan` gave an Error if `meas` had a time vector of length zero. This manifested itself as an issue when a component_measurement comes from a file without any data, as one has when starting a biologic experiment with more steps than you actually end up using and read it in with `read_set` (Issue #93). 

What it doesn't fix is whether a cut measurement should have a list of cut component measurements. I apparently didn't feel great about it when writing the code 2.5 years ago: https://github.com/ixdat/ixdat/blob/b0cef164f410daa44aea7903e07a887645bf879b/src/ixdat/measurements.py#L1071


AnnaWiniwarter (approved):
Solves issue #93 also locally with the dataset that initially prompted me to raise the issue.

I do not have a qualified opinion on whether a cut measurement should have a list of cut component measurements - if you think this is important, maybe raise a separate issue?

AnnaWiniwarter:
Thanks for the explanation. What do you mean by "want" just the MS data though? I can imagine that I might want to plot only the MS data of a specific cycle without showing EC data. 
I think the method I wrote was removed/changed because it was kind of misusing ixdat, but for example, I wanted to use EC cycles to "mark" when I change gas composition for making a gas calibration curve, so I can automatically select the relevant periods in the MS data rather than having to manually select the tspan every time. If that makes sense. So then I wouldnt have any EC data, except cycles, but would need that for analysing the MS data.

ScottSoren:
"want" could mean for export, for plotting, etc, for example:
```
ec_ms = ec + ms
part = ec_ms.select(cycle_number=3)
ms_part = [m for m in part.component_measurements if m.technique == "MS"][0]
ms_part.export("mass spec data during cycle 3.csv")
```
I'll leave it in for now, can always be removed if there's no evidence anyone uses it and it causes problems in the future.

PR #138, opened 2023-11-12
Zilien spectromeasurement

### LINKED ISSUE #117: Zilien spectra reader is not working
There is a spelling mistake in the data exported by Zilien (`"Mass  [AMU]"` with a double space), while the reader expects this to be `"Mass [AMU]"` (single space). 
Also, when preparing the dict: it needs to be `"name": self.path_to_spectrum.name` rather than `"name": path_to_spectrum.name`

Note that I'm using the quant_integration branch, maybe this has been fixed elsewhere?

ScottSoren:
This PR enables combined reading of zilien measurements and spectra. It'll close #117 

First, it improves the calling of readers:
`Measurement.read(my_MID_file, reader="zilien")` calls `ZilienTSVReader.read()` and returns a `Measurement`. Specifically, a `MSMeasurement`, `ECMSMeasurement`, or `SpectroMSMeasurement` based on what data it finds.
 `Spectrum.read(my_mass_scan_file, reader="zilien")` calls `ZilienSpectrumReader.read()` and returns a `MSSpectrum`. (previously the reader had to be called "zilien_spec".
 
 Then, it adds power to `Measurement.read(my_MID_file.tsv, reader="zilien")` by including the spectra if there is a "mass scans" folder associated with "my_MID_file.tsv". The way it finds it is by removing the first 20 characters and extension from the file name and adding "mass scans" to get the folder name. This is fragile, but that's because Zilien's way of saving data is fragile IMO (@KennethNielsen). I would prefer for zilien to include a list of auxiliary files in the header, as ixdat's `SECExporter` does. I still need to set up a `SpectroMSMeasurementExporter` which does the same. I'll open that as an Issue. Let me know if you want exporting and re-reading to work as part of this PR.
  
 If the user hasn't turned off the feature by passing `include_mass_scans=False` to the `read()` and the reader finds mass scans, it returns a `SpectroMSMeasurement`. This has a `spectrum_series` attribute which can be indexed to get the individual spectra. It also plots with a heat plot in one panel and MID data in the other panel. Getting this plot to look right involved some unexpected subtle debugging to get the time axes correctly lined up, but it should be robust now! (@AlexanderKrabbe , this would also have affected your previous `SpectroMSMeasurement` plots.) A remaining problem with the plots is that the `heat_plot` function assumes that the mass scans were taken evenly spaced in time. (Unfortunately, for zilien measurements, this is not generally the case.) I'll open this as a separate Issue as well.
 
The usage is described more succinctly in NEXT_CHANGES.rst
 I've used the files here for demo with the development_scripts/demo_zilien_spectrum_reader.py`: https://www.dropbox.com/sh/5elp3qz7wkfk68y/AACJj0rYH9bVZcxSZK9pOHk_a?dl=0
 Gladly try it out on your own zilien data with mass scans! :)


KennethNielsen:
I will ask @matenestor to review this in my place.

AnnaWiniwarter (changes requested):
Had a look at this and tested it with a different dataset - seems like it's working fine, but I'm not happy merging this without the heatmap part being implemented well (it makes it a bit useless IMO). 
In my dataset, I have three spectra recorded in the same gas, before and after purging with another gas. This is what the plot looks like now:
image
I.e. it shows the heatmap as if the spectrum was the same also over the timespan where the gas composition was actually quite different.
This is unfortunately not very useful like this. (I didnt look into how this plot is made, so I don't have any suggestions for how to improve it, but I do think it's a bit of an issue)

ScottSoren:
Thanks for trying it out! 
Can you describe what you would like the plot to look like for this dataset?

It could be that this dataset just shouldnt have the spectra plotted by default? It's not ixdat's job to make sure that the data makes sense, and if your spectrum sample rate is way slower than the changes taking place, there's no fixing that. 
I think a PVMS dataset with some masses tracked would look mych better, if you have one of those available? (@matenestor, I think Zilien really needs a take-spectrum-every-X-minutes function.) 

Maybe ixdat should, for this dataset, import the spectra (to enable `ms_meas[0].plot()`) and do the default MS plot just with an annotation indicating where the spectra are taken?

AnnaWiniwarter:
> It's not ixdat's job to make sure that the data makes sense, and if your spectrum sample rate is way slower than the changes taking place, there's no fixing that

I agree, ixdat's job is not to make sure the data makes sense, but it should not show data that is not actually measured. I suppose, if you do record specta repeatedly with a high time resoltion (compared with the overall experiment time at least), it makes sense to just plot a continuous heatmap like this. Just as it's ok to draw a line between the individual datapoints of the MID data, since plotting them point-by-point would look the same anyway (thinking about this now, plotting them as points might actually make it easier to distingish spikes from peaks, but this is definitely out of scope here).

In the case of manually triggered spectra at random times of the experiment, however, it's simply not accurate to plot the heatmap also in between the spectra. The most straightforward would be to just do as you suggest and not plot the heatmap by default.  
Alternatively (and the way I think it should behave) is it should just show the heatmap for the timespan where the spectrum was actually recorded, i.e. as a thin line. Does the spectrum know how long time it took to record it? that would give the thickness of the line (thickness on the time scale, if that makes sense). If you record spectra frequently, the lines will be closer together, possibly with little distance in between. If you don't, there will just be few lines on a time axis. This would be the most robust way for all types of spectra heatmap vs time plots IMO.

AnnaWiniwarter:
I agree, Zilien needs an option to automatically record a spectrum every x amount of time. But ixdat has to have a way of dealing with the data we get from our instrument software, however imperfect that might be.

matenestor:
The "option to automatically record a spectrum every x amount of time" sounds like very useful feature! Thanks for mentioning it. I will add it to our backlog.

Regarding Ixdat plotting data that weren't measured -- the connected spectra, it sounds like a bug you caught in this PR. It can be fixed. I am commenting on one line in the review that is probably causing it.

And regarding the review, there is only one file remaining that I would like to look at more and then I will send the review. After that I would like to experiment a bit with this feature as well. It looks good in general.

src/ixdat/spectra.py
```diff
@@ -647,6 +667,7 @@ def spectrum_series(self):
         """The `SpectrumSeries` with the spectral data"""
         if isinstance(self._spectrum_series, PlaceHolderObject):
             self._spectrum_series = self._spectrum_series.get_object()
+        self._spectrum_series.tstamp = self.tstamp
```
matenestor, line 670: Why should the `tstamp` be reassigned every time this property is used?

src/ixdat/spectra.py
```diff
@@ -508,6 +508,26 @@ def from_spectrum_list(cls, spectrum_list, **kwargs):
...
+            self._field = Field(
+                name=self._field.name,
+                data=self._field.data,
+                unit_name=self._field.unit_name,
+                axes_series=[
+                    time_shifted(self._field.axes_series[0], tstamp=self.tstamp),
+                    self._field.axes_series[1],
+                ],
+            )
+        return self._field
```
matenestor, line 529: I think there is a case here when this method could return `None` when `self._field` is not a `PlaceHolderObject` and when the first axis has different timestamp than the one in the object.
Also, if `tstamp`s are `floats` they shouldn't be compared with `==`.

matenestor (changes requested):
I commented on one line where there might be a problem causing connected mass scan plots. And I would like to discuss more the `_field` property.

src/ixdat/spectra.py
ScottSoren: The `SpectrumSeries` returned here should have the same `tstamp` as the `Measurement`, so that the `Spectrum` objects it contains have `spectrum.t` with respect to the `Measurement.tstamp` so that anyone with a plot of the measurement can identify when the spectrum was taken. The `Measurement`'s tstamp can be reset by the user at any time by e.g. `meas.tstamp += 1000` to switch the absolute time at which relative t=0. 
The proper way to meet these requirements, I think, is to make `Measruement.tstamp` a property with a setter and a getter, the setter of `SpectroMeasurement` being modified to update `self._spectrum_series.tstamp`. I just liked `tstamp` being a simple attribute, so took this lazy solution.
I will fix it, doing it the "right" way just described instead.

src/ixdat/spectra.py
ScottSoren: Nice catches! 

I guess that conditional should instead be 
`if np.abs(self._field.axes_series[0].tstamp - self.tstamp) > tolerance:`
But I don't know what `tolerance` should be? It could be widely different depending on the method. Maybe
```
t = self._field.t    # shortcut to self._field.axes_series[0].data
tolerance = 1/100 * (t[-1] - t[0]) / len(t)`
```
?

I think this will never return `None`. `self._field` can only be a `Field` or a `PlaceHolderObject`. If it is a `Field` which already has the same `tstamp` in its axes series as the `Measurement`s tstamp, then it can be returned without modification.

src/ixdat/plotters/ms_plotter.py
```diff
@@ -647,6 +646,9 @@ def plot_measurement(
             vmax=vmax,
         )
 
+        # Get the time variables aligned!
+        ax_heat_plot.set_xlim(axes[ms_axes].get_xlim())
+
```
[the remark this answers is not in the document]
ScottSoren: Yes! 
I need to rewrite this plotting function to either put in zeros where spectra weren't taken and/or interpolate (`imshow` is too dumb for that). And I need to write a new plotting function to be the default `plot()` for zilien-read `MSSpectroMeasurment`s which does something [like what @AnnaWiniwarter suggested](https://github.com/ixdat/ixdat/pull/138#issuecomment-1854206684).

src/ixdat/spectra.py
ScottSoren: I've implemented the above solution, calling it `t_tolerance`. Let me know what you think!

src/ixdat/spectra.py
ScottSoren: Done.
Turns out `tstamp` was already a protected property :) 
So very little to add.

ScottSoren:
@AnnaWiniwarter , this is now the result of 
`zilien_meas_with_spectra.plot()`

Boring data, but let me know if the format looks like what you meant!
 
`zilien_meas_with_spectra.spectrum_series.continuous = True` switches it back to the old behavior.

@matenestor , that took a fair bit of new code - if you have time for another code review, that would be great!
demo

ScottSoren:
@matenestor , here is a diagram of all the classes having to do with spectrum series. There's a brief description of the justification for each class. I'm not sure it's optimal, and open to revising the design!
as png for rendering in this comment: 
ixdat class maps
as pdf for looking good with clickable links: [ixdat class maps.pdf](https://github.com/ixdat/ixdat/files/13714424/ixdat.class.maps.pdf)
as pptx for editing

AnnaWiniwarter:
> @AnnaWiniwarter , this is now the result of `zilien_meas_with_spectra.plot()`
> 
> Boring data, but let me know if the format looks like what you meant!
>
@ScottSoren, this looks great (and also works as I had in mind with my data)! :) thank you!

AnnaWiniwarter (approved):

src/ixdat/spectra.py
matenestor: Maybe [numpy.is_close()](https://numpy.org/doc/stable/reference/generated/numpy.isclose.html) would be sufficient?

Saying that something "will never happen" in programming means that it is going to happen in a very weird edge case or while adding a new feature. :smile: Now it looks it will always contain at least a `PlaceHolderObject`, so we can have a look later.

matenestor (commented):
I have a question about retrieving a field from a spectrum series object after a tstamp changes.

src/ixdat/spectra.py
```diff
@@ -505,9 +522,31 @@ def from_spectrum_list(cls, spectrum_list, **kwargs):
...
+        if abs(self._field.axes_series[0].tstamp - self.tstamp) > self.t_tolerance:
+            self._field = Field(
+                name=self._field.name,
+                data=self._field.data,
+                unit_name=self._field.unit_name,
+                axes_series=[
+                    time_shifted(self._field.axes_series[0], tstamp=self.tstamp),
+                    self._field.axes_series[1],
+                ],
+            )
```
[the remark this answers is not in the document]
ScottSoren: Not sure I completely understand. I think the `t_tolerance` should not change when `tstamp` is changed. `t_tolerance` is related to the time spacing between the spectra, which doesn't change. `tolerance_by_spacing`, to which `t_tolerance` defaults, is likewise unchanged, since it is the time spanned by the `tseries` (i.e., `t[-1] - t[0]`), which doesn't change when `t` is shifted by a constant.

src/ixdat/readers/zilien.py
```diff
@@ -69,7 +70,8 @@ def parse_metadata_line(line):
     if type_as_str == "string":
         return full_name, value
     elif type_as_str == "int":
-        return full_name, int(value)
+        # Zilien writes some ints as floats, e.g. "0.0".
+        return full_name, int(float(value))
```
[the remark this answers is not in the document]
ScottSoren: In python 3.11 on ubuntu:
```
>>> test = "0.0"
>>> int(test)
ValueError: invalid literal for int() with base 10: '0.0'
>>> float(test)
0.0
>>> int(float(test))
0
```

src/ixdat/readers/zilien.py
```diff
@@ -589,21 +626,46 @@ class ZilienSpectrumReader:
...
+        with open(self.path_to_spectrum, "r") as f:
+            in_header = True
+            i = 1
+            metadata_dict = {}
+            while in_header:
+                name, value = parse_metadata_line(f.readline())
+                metadata_dict[name] = value
+                if i == metadata_dict.get("num_header_lines", 0):
+                    in_header = False
+                i += 1
```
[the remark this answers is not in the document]
ScottSoren: renamed `i` to `num_line`.

KennethNielsen:
> I agree, Zilien needs an option to automatically record a spectrum every x amount of time. But ixdat has to have a way of dealing with the data we get from our instrument software, however imperfect that might be.

I'll record this as a feature request. As soon as Scott buys an ECMS as well, we will have two customer requests for it, which should be enough to just implement it ;)

KennethNielsen:
Looking at the class hierarchy, I think it would be a good time to have a new design discussion soon, as the inheritance and the number of classes is getting more and more complicated. Although we should not delay this PR of course. I will see if I can work up and proposal in a presentation along with @matenestor and then we can see if it shouldn't possibly fit in a separate session.

matenestor (approved):
Good, approved

ScottSoren:
Merged! :) Thanks, @AnnaWiniwarter and @matenestor for the constructive reviews of this one!

Issue #79, opened 2022-05-23
Use a background found elsewhere in calibration functions


ScottSoren:
Right now the only way to define a background value in calibration functions like `MSInlet.gas_flux_calibration` is by using `tspan_bg`. This means that there has to be a background signal measurement in the same `MSMeasurement` as the calibration. However, what if I have the background value from elsewhere? I might want to do this:

```
M2_bg = np.mean(my_background_measurement.grab("M2")[1])
cal_H2_M2 = my_chip.gas_flux_calibration(my_real_measurement, mol="H2", mass="M2", bg=M2_bg, ...)
```
where the key-word argument here called `bg` is not presently implemented.

AnnaWiniwarter:
I tried to look into this to see if I can take care of this when adding a `gas_flux_calibration_curve()` (issue #80) (I might, but it's not as directly related as I hoped)
- do I understand correctly that `set_bg()` is like a calibration for a set of data that will overrule any tspan_bg you pass basically do any function that then uses grab() to get the signal?

ScottSoren:
Hi Anna,
Having called `set_bg` _does not_ override use of `tspan_bg` in other functions. It just sets a "default background" to be subtracted when background subtractied or implied and no `tspan_bg` is given.
By "implied" here I mean when the plotting function thinks that the user wants a background subtracted. Code is more precise than words, so here's that bit in `ECMSPlotter.plot_measurement`:
```
        if remove_background is None:
            remove_background = not logplot
```
So the user can request the default (preset with `set_bg`) background be removed if and only if the user calls the function with `remove_background=True` (by default it is `None`) OR `logplot=False` (by default it is `True`). Explicitly saying `remove_backgorund=False` also turns it off whether or not a logplot is requested.

An important bit is that the background determined by `set_bg` carries over when you chop up the measurement (for example by indexing an `ECMSCyclicVoltammogram` to select a cycle), even if that means losing the section with the tspan used for the original `set_bg`.

I noticed a problem while testing `set_bg` now to make sure I answer right: there's no way to unset it. `reset_bg()`, which should do this, seems to be broken. I'll open a different issue for this.

However, `set_bg` does not solve the original Issue here, becuase it can also only define a background from a `tspan_bg`. It does not give a user a chance to manually define the background signal.

AnnaWiniwarter:
Hi Soren, 
Thanks for the thorough explanation.
[...]

PR #154, opened 2023-12-22
Background class

AnnaWiniwarter:
@ScottSoren and I implemented a background-class based on the calibration class, as discussed.

- We left the option `"tspan_bg" `which will subtract a constant background as a possibility in `grab()`. 
   Rationale to that decision: 
-- It's the simplest possible background
-- We might want to do that on top of a more sophisticated background anyway. 
-- It would require a lot of refactoring to remove the option without clear benefit.

- So far, one type of background is implemented: The class `MatrixInterferenceBackground,` which allows to subtract the signal of a matrix mass (eg solvent) multiplied by a factor determined where the only contribution to M is from the matrix molecule.
- The background subtraction works both for masses and mols (i.e. calibrated and uncalibrated data)
- This is a first start to solving #79 

We need to decide:
a) do we merge this now and then discuss whether there should be a common base class for calibration, background and filter, or should we refactor now and merge later? (I am pro merging now)
b) define other background classes before merging? (Eg some linear between points, ...)

Did I forget anything, @ScottSoren?

### DIFF
--- a/NEXT_CHANGES.rst
+++ b/NEXT_CHANGES.rst
@@ -20,5 +20,11 @@ ixdat 0.2.9
 techniques
 ^^^^^^^^^^
 
+``MSMeasurement.remove_matrix_interference`` associates a background with the measurement
+  which subtracts a ratio (measured when there's only solvent) times the signal of the primary solvent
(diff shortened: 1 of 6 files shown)


ScottSoren:
Archiving pending resolution of #164

Issue #158, opened 2024-01-24
Problem when trying to add two SpectroMSMeasurement objects


AnnaWiniwarter:
I came accross a problem before when importing data since `SpectroMSMeasurement` was merged (we discussed this on Slack). Now I ran into it again, and it seems to be a problem when trying to add something to a `SpectroMSMeasurement`  object (I tried adding a second `SpectroMSMeasurement` or a plain `MSMeasurement`.  I suppose this is not expected and a bug? @ScottSoren?

![image](https://github.com/ixdat/ixdat/assets/96479350/26dde742-14bd-4a02-bea9-91a3592e2ba1)

matenestor:
I am getting this error as well when I am running external tests. The problem is in `Measurement.from_dict()` where `measurement` variable is not initialized, but it is trying to be used. [Lines 225-238](https://github.com/ixdat/ixdat/blob/e66675fc4aa81c75d6e857983e5a87434c042786/src/ixdat/measurements.py#L238)

There is just a `print` in the `except` block, but it should be `raise` which would stop the program, because it failed to create a new object. And it is not a user's problem, so I am in doubt whether the message should be there in this case.

In Anna's case, this method is used internally by ixdat while creating a new object from the two. That is one thing. Then it also complain that `SpectroMeasurement` must be initialized with `spectrum_series` and `spectrum_id`, which are probably not provided when creating the `SpectroMeasurement`.

In my case, it is the tests, because some of the datasets contain mass scans and then it fails, because of [this spectra addition](https://github.com/ixdat/ixdat/blob/e66675fc4aa81c75d6e857983e5a87434c042786/src/ixdat/readers/zilien.py#L239). The exception I am getting is: `"Measurement.__init__() got an unexpected keyword argument 'ec_technique'"`

---

Looks like there is a lot of fixes to be done plus @KennethNielsen the CI is not working as intended, because it should be running the external tests and this problem would had been caught already before.

KennethNielsen:
I agree there are obviously bugs here. Adding two `SpectroMSMeasurement`s should probably be allowed (maybe under certain constrictions), but I'm unsure if we ever decided what should happen if you add a `SpectroMSMeasurement` with a `MSMeasurement` (since they are sort of orthogonal in nature). I think this is a question that @ScottSoren already raised, but I don't think we made a decision.

[...]

PR #163, opened 2024-01-31
Run external tests in CI, closes #162

### LINKED ISSUE #162: CI not running tests based on external data
See the comment here: https://github.com/ixdat/ixdat/issues/158#issuecomment-1918616372

KennethNielsen:
This PR is meant to enable external tests in CI. This part is already setup, but some tests are failing as mentioned here https://github.com/ixdat/ixdat/issues/158#issuecomment-1918616372. But, since https://github.com/ixdat/ixdat/pull/159 is also touching the external tests, I will wait updating this one with ignoring the tests that are still failing.

EDIT. #159 is now merged. I updated this one, so the commit of the submodule corresponds to the latest data. I also marked all the tests that fails as "skip" for now and will leave fixing those in the capable hands of @matenestor and @ScottSoren. @matenestor the skip reason includes the link to issue 158, so that are easy to find.

But this PR accomplished its main purpose, which is to make sure that externals tests are run in CI.

NOTE: I accidentally left a completely broken submodule setup, when I originally made the large data files sub module. The problem was that it was added as a submodule in the config but never committed. This was the reason why nothing happened when you ran `git submodule update --init --recursive`. This is now fixed and the submodule should work as intended.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "submodules/ixdat-large-test-files"]
 	path = submodules/ixdat-large-test-files
-	url = https://github.com/ixdat/ixdat-large-test-files.git
+	url = https://github.com/ixdat/ixdat-large-test-files
--- a/tests/functional/test_eclab_parser.py
+++ b/tests/functional/test_eclab_parser.py
@@ -10,7 +10,7 @@ from pathlib import Path
 
 from ixdat import Measurement
 
-from pytest import approx, fixture, mark
+from pytest import approx, fixture, mark, skip
 
 
 DATA_DIR = Path(__file__).parent.parent.parent / "submodules/ixdat-large-test-files/"
@@ -303,7 +303,6 @@ TEST_DATA = {
 @fixture(scope="module", params=tuple(TEST_DATA.items()))
 def measurements_with_data(request):
     """Load all measurements from files and connect it with the test data."""
-
     filename = request.param[0]
     columns_data = request.param[1]
     measurement = Measurement.read(DATA_DIR / filename, reader="biologic")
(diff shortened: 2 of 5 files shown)


.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "submodules/ixdat-large-test-files"]
 	path = submodules/ixdat-large-test-files
-	url = https://github.com/ixdat/ixdat-large-test-files.git
+	url = https://github.com/ixdat/ixdat-large-test-files
```
matenestor, line 3: I was experimenting with adding a submodule url with `.git`suffix and without it, and it worked both ways. However, the url was consistent with the one in the `.git/config` file. Are you sure this will not cause problems?

matenestor (approved):
I fetched your branch and updated the submodules. It works as expected now. I have one comment with a small question, otherwise it is ready to go.
The tests will need fixing, but that is work for another PR, since they complain about multiple places.

ScottSoren:
I'll make my PR today to here so that it passes the tests :)

tests/functional/test_eclab_parser.py
```diff
@@ -318,6 +317,15 @@ def test_shape(measurements_with_data):
...
 
     """
+    if measurements_with_data[0].name in (
+        "multiple_techniques_dataset_01_02_CVA_C01.mpt",
+        "multiple_techniques_dataset_01_07_ZIR_C01.mpt",
+        "multiple_techniques_dataset_01_08_CVA_C01.mpt",
+        "dataset_with_loop_01_01_OCV_DUSB0_C01.mpt",
+        "dataset_with_loop_01_02_CVA_DUSB0_C01.mpt",
+    ):
+        skip("TEMP SKIP. BROKEN, SEE: https://github.com/ixdat/ixdat/issues/158")
```
ScottSoren, line 327: Wait a second @KennethNielsen , these have to to with .mpt reading, so not related to #158. What goes wrong here?

PR #165, opened 2024-02-06
Unskip biologic tests

matenestor:
Fixes for failing tests for reading biologic files and checking dataset integration consistency. This is a patch after #134 which now creates `cycle number` and `Ns` series with `=0` suffix when they are missing in an .mpt file. (This triggered a change in Zilien code as well :smile: ).

### DIFF
--- a/tests/functional/test_zilien_reader.py
+++ b/tests/functional/test_zilien_reader.py
@@ -186,7 +186,6 @@ def datasets(request):
     return tsv, mpts, mpt_time_offsets
 
 
-@pytest.mark.skip("TEMP SKIP. BROKEN, SEE: https://github.com/ixdat/ixdat/issues/158")
 @pytest.mark.external
 @pytest.mark.parametrize(
     ["cls_or_technique", "expected_series"],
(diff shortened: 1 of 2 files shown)


matenestor:
Is a changelog entry required for this?

tests/functional/test_zilien_reader.py
```diff
@@ -259,6 +260,13 @@ def test_series_match(self, datasets):
             for mpt_name in mpt.series_names:
                 if mpt_name == "time/s":
                     assert f"Biologic {mpt_name}" in tsv.series_names
+                elif mpt_name.endswith("=0"):
+                    # ConstantValue series with "=0" suffix is created when an
+                    # essential series is not in the parsed dataset.
+                    # This suffix is removed by Zilien when integrating .mpt files,
```
ScottSoren, line 266: @matenestor , what is Zilien's behavior when running with EC data incorporated from a Biologic file with `Ns`? Without `Ns`?
When does Zilien add an Ns column to the data? Does it remove a "=0" suffix, or just (in contrast to ixdat) not add such a suffix?

ScottSoren (approved):
One question but looks good! Thanks for diagnosing and fixing these tests :)

tests/functional/test_zilien_reader.py
```diff
@@ -206,10 +205,15 @@ def test_read(cls_or_technique, expected_series):
     if isinstance(cls_or_technique, str):
         expected_measurement_class = TECHNIQUE_NAME_TO_CLASS[cls_or_technique]
         measurement = Measurement.read(
-            PATH_TO_DATAFILE, reader="zilien", technique=cls_or_technique
+            PATH_TO_DATAFILE,
```
ScottSoren, line 208: This doesn't seem like the most elegant way to do it... 
what do we want the behavior to be when calling `ECMeasurement.read(path_to_tsv)" on a .tsv with associated mass scans? 
I guess in that case it should ignore the mass scans, right? I'll make the suggestion in my PR.

ScottSoren:
I think no Changelog needed for this. Merging (but see questions in comments). Thanks for the PR!

tests/functional/test_zilien_reader.py
matenestor: @ScottSoren Zilien will now remove the "=0" suffix created from all columns names containing it, so they are there in a pure form, since ixdat doesn't load the columns as `ConstantValue` anyway.
Though, there is a column `raw_current=0` (at least this one) being produced by Zilien, because back then I didn't think there would be a difference. But I am patching the dataset generation now.

PR #167, opened 2024-02-18
Nordic Electrochemistry tdms

ScottSoren:
A reader for the .tdms files made by the labview-based EC4-DAQ software by Nordic Electrochemistry for their potentiostats.
It makes use of the [npTDMS package](https://nptdms.readthedocs.io/en/stable/).

(If #166 is merged first, this branch should be rebased and merged to main.)

### DIFF
--- a/NEXT_CHANGES.rst
+++ b/NEXT_CHANGES.rst
@@ -12,5 +12,11 @@ links to relevant Issues, Discussions, and PR's on github with the following for
 
 `PR #XX <https://github.com/ixdat/ixdat/pull/XX>`_
 
-For ixdat 0.3.0
+For ixdat 0.2.9
 ===============
+
+readers
+^^^^^^^
+
+- The ``NordicTDMSReader`` (reader="nordic") has been added for reading the .tdms files
+  produced by Nordic potentiostat. So far it just reads current, potential, and impedance.
--- a/development_scripts/reader_demonstrators/demo_nordic_tdms_reader.py
+++ b/development_scripts/reader_demonstrators/demo_nordic_tdms_reader.py
@@ -4,8 +4,14 @@ from ixdat import Measurement
 
 data_dir = Path.home() / "Dropbox/ixdat_resources/test_data/nordic_tdms/24B07_0_Pt"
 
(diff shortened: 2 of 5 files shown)


ScottSoren (commented):
Additional notes from Gustav:
- call tstamp "unix_time", at least in export.
- Make a human-readable timestamp (iso 8601)
- Overload some methods to allow multiple units, eg A_el_m2 in ECMeasurement.calibrate (though ideally SI should be default and others like cm^2 should require subsctipt)
- Make default symbol for potential configurable (most electrochemists prefer "E" for potential)
- use "positive" and "negative" instead of "anodic" and "cathodic" when talking about scan direction (probably only affects docstrings)

FrederikLizakJohansen:
I've made the following changes to the TDMS reader class:

- Units are now read from the TDMS channel properties (`unit_string`) rather than hardcoded. Current is converted from A to mA to follow ixdat convention.
- `Phase_E` unit is now read from the file (`rad`, radians) instead of `None`.
- ISO 8601 `datetime` is included in the measurement metadata.
- Macro file parsing: if a `*.EC_Macro` file is present in the same directory as the `.tdms` file, it is automatically parsed and included in the metadata as a list of experiment steps (hardware control, gas, CV, hold parameters). The user is warned if no macro file is found, or if multiple are found.
- Minor fixes: removed unused import, replaced `print()` with `warnings.warn()`, fixed a scope bug where `tseries_EIS` could be undefined when `Phase_E` was present without `Z_E`.

The following points raised in the review are left for a future PR as they concern the broader ixdat API:
- Renaming `tstamp` to `unix_time` in exports
- Supporting multiple units in `ECMeasurement.calibrate` (e.g. `A_el_m2`)
- Making the default potential symbol configurable
- Replacing "anodic"/"cathodic" with "positive"/"negative" in docstrings

FrederikLizakJohansen:
CI tests for python 3.7 are currently failing due to unrelated issues, which are being addressed separately in PR https://github.com/ixdat/ixdat/pull/194

FrederikLizakJohansen:
Everything else look great! 
Resolving the conflict with NEXT_CHANGES.rst and then merging 🎉

PR #171, opened 2024-03-11
Fix multical to work without passing tspan_bg

AnnaWiniwarter:
@ScottSoren I just made the tiny change in multical to not subtract the entire measurement if no tspan_bg is passed. Decided to pull request this now since multical seems to be working fine (our issues seem to be caused by something else).
I didnt add anything to next changes. let me know if I should.


ScottSoren (commented):
Looks good to me! Merging.

PR #176, opened 2024-05-08
ms_plotter works

AlexanderKrabbe:
ms_plotter works

ms_plotter.vs() needs some clean up

The main thing is how pint lets you set a matplotlib formatter for the units, so when axes are passed on, the label dynamically updates if the unit of the axes shoul change:

ax = meas_ms.plot()
ax.yaxis.get_unit() -> "A"
ax.yaxis.set_units(ureg.nA)
plt.plot()

data is correctly plotted

There is a nice thing about pint with their class Context.

This class could be a nice way to provide context for the measurement when units changes from mol/s to mol/s/cm^2

### DIFF
--- a/src/ixdat/plotters/ms_plotter.py
+++ b/src/ixdat/plotters/ms_plotter.py
@@ -130,8 +130,8 @@ class MSPlotter(MPLPlotter):
             if logplot:
                 try:
                     v.m[v.m < MIN_SIGNAL] = MIN_SIGNAL
-                except DimensionalityError:
-                    v[v < MIN_SIGNAL.m] = MIN_SIGNAL.m
+                except (DimensionalityError, AttributeError):
+                    v[v < MIN_SIGNAL] = MIN_SIGNAL
                     
             if logdata:
                 logplot = False
@@ -152,8 +152,6 @@ class MSPlotter(MPLPlotter):
             # expect always to plot against time
             #x_unit_factor, x_unit = [1, ureg(x_unit).u] if use_quantity else self._get_x_unit_factor(x_unit, DEFAULT_UNIT["time"])
             
-
-            
             ax.plot(
                 t, #* x_unit_factor,
                 v * unit_factor,
@@ -161,6 +159,7 @@ class MSPlotter(MPLPlotter):
                 label=v_name,
                 **kwargs,
             )
+        
         
         if logdata:
             ax.set_ylabel(ylabel)
@@ -478,10 +477,11 @@ class MSPlotter(MPLPlotter):
             ax = (
                 axes[0]
                 if axes
-                else self.new_ax(ylabel=f"signal / [{DEFAULT_UNIT['signal']}]", 
-                                 xlabel=f"time / [{DEFAULT_UNIT['time']}]",
-                                 use_quantity=use_quantity)
+                else self.new_ax()#ylabel=f"signal / [{DEFAULT_UNIT['signal']}]", 
+                    #xlabel=f"time / [{DEFAULT_UNIT['time']}]",
+                #             use_quantity=use_quantity)
             )
+
         # as the next simplification, if they give two things (v_lists), we pretend we
         #   got one (v_list) but prepare an axis for a recursive call of this function.
         if v_lists:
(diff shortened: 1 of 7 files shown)


src/ixdat/plotters/ms_plotter.py
```diff
@@ -119,25 +127,69 @@ def plot_measurement(
...
+                    v * ureg.mol / ureg.s if quantified else v * ureg.A
+                ),  # ixdat internal default is A and mol/s for un/calibrated signals
                 color=color,
                 label=v_name,
                 **kwargs,
             )
-        ax.set_ylabel(f"signal / [{unit}]")
-        ax.set_xlabel(f"time / [{x_unit}]")
+
+        if logdata or (use_quantity and v.check(ureg.dimensionless)):
```
ScottSoren, line 189: Please break this up into `if` + `elif` so the logic is easier to follow.

src/ixdat/plotters/ms_plotter.py
```diff
@@ -242,14 +316,29 @@ def plot_vs(
...
         unit = specs_this_axis["unit"]
-        unit_factor = specs_this_axis["unit_factor"]
+        # unit_factor = specs_this_axis["unit_factor"]
+
+        t, x = measurement.grab(
+            x_name, tspan=tspan, include_endpoints=True, return_quantity=use_quantity
+        )
+
+        if use_quantity and x_unit and not x.check(x_unit):
+            error_message = (
```
ScottSoren, line 332: see previous comment on this. 
Also is there a way to avoid repeating such similar code?

src/ixdat/plotters/ms_plotter.py
```diff
@@ -266,35 +356,77 @@ def plot_vs(
                     tspan_bg=tspan_bg,
                     remove_background=remove_background,
                     include_endpoints=False,
+                    return_quantity=use_quantity,
+                )
+
+            if ureg("mol / s / cm ** 2").check(unit):
```
ScottSoren, line 362: Is there a way to do somethign like
`if "length^{-2}" in unit.dimensions: ...`
? 
Then there wouldn't have to be a repeat of this block for quantified vs non-quantified.

src/ixdat/plotters/ms_plotter.py
```diff
@@ -266,35 +356,77 @@ def plot_vs(
...
+                logplot = False
+                vu = v.to(unit).u if use_quantity else unit
+                v = (
+                    np.log(v.to(unit).m) * ureg.dimensionless
+                    if use_quantity
+                    else np.log(v / (1 * unit).to_base_units().m) * ureg.dimensionless
+                )
+                ylabel = f"{vu:~ixdat_log}"
+
             x_mass = np.interp(t_v, t, x)
```
ScottSoren, line 406: I know this isn't you but `x_mass` is a wierd name... looks like it should be called `x_v` (x for the specific value being plotted) to be consistent with the nomenclature `t_v`.

src/ixdat/plotters/ms_plotter.py
```diff
@@ -321,55 +453,23 @@ def plot_vs(
...
-            }[new_x_unit]
-        except KeyError:
-            warnings.warn(
-                f"Can't convert original unit '{original_unit_name}' to new unit "
-                f"'{new_x_unit}'. Plotting using original unit '{new_x_unit}' with "
-                "unit_factor=1 (one).",
-                stacklevel=2,
+            ax.xaxis.set_units(x_unit)
+
+        if unit and not logdata:
```
ScottSoren, line 464: Again, there must be a way to avoid this...

ScottSoren (changes requested):
Hi @AlexanderKrabbe , 

Great work. I like a lot that the manual unit conversions disappear from the plotting modules. 

However, I'm surprised to see so much more code get added than removed. That indicates that pint units isn't giving us the desired simplification of ixdat yet. 
Most of my comments are pointed towards that. I think a lot of the stuff added to units.py is not needed, but on the other hand some switches in units.py or data_series.py controlled by a variable defined in `config.plugins` could make it all work. 

Another problem is that the development scripts that I have to test this on reactor data, **demo_ixdat_reader_TPMS.py** and **demo_ms_spectrum_measurement.py** don't work yet, presumably because the readers don't correctly set the units. Could you take a look at those and get them working? Let me know if you don't have the data.

Sorry this review comes so slow! No need to implement all my comments - just respond with your thoughts of how to proceed, unless a simplifying implementation is clear enough to do quickly. I don't think any substantial changes should be made which don't consist of a net decrease in code! 

When I get your response to the comments, I'll merge it onto the units branch and we can take it from there.

PR #177, opened 2024-05-15
Hydrogen isotope cals

AnnaWiniwarter:
To streamline calibration with hydrogen isotopes from hydrogen evolution in D2O, I added the following:
1) option to add an `isotope_factor` to correct the flux calculated from EC (based on binomial distribution and knowing/being able to determine the ratio of H/D in the electrolyte. for simplicity, this is done outside the method, isotope_factor is literally just a multiplication factor for the calculation of n_dot from EC)
2) standard colors for H2@M2, H2@M3 and H2@M4
4)` integrate_flux()` analogous to` integrate_signal()` (I was quite surprised that this didnt exist yet tbh)

Considering how minor the changes are and that I would like to share a script using those with not-so-ixdat-savy users, I would appreciate a quick merge and release of a new ixdat version. :)

PS: I tried to format with Autopep8, but it seems it's still failing checks :( sorry!


src/ixdat/techniques/ec_ms.py
```diff
@@ -72,14 +72,16 @@ def as_cv(self):
...
-    def ecms_calibration(self, mol, mass, n_el, tspan, tspan_bg=None):
+    def ecms_calibration(self, mol, mass, n_el, tspan, isotope_factor=1, tspan_bg=None):
         """Calibrate for mol and mass based on one period of steady electrolysis
 
         Args:
             mol (str): Name of the molecule to calibrate
             mass (str): Name of the mass at which to calibrate
             n_el (str): Number of electrons passed per molecule produced (remember the
                 sign! e.g. +4 for O2 by OER and -2 for H2 by HER)
+            isotope_factor (str): Factor to multiply EC flux by to deconvolute
```
ScottSoren, line 83: Could we just call this faradaic_efficiency? 
With a comment in this docstring that it also works for isotopes.

src/ixdat/techniques/ec_ms.py
```diff
@@ -296,6 +305,8 @@ def siq_ecms_calibration(self, mol, mass, n_el, tspan, tspan_bg=None):
...
```
ScottSoren, line 308: ditto. faradaic_efficiency.

src/ixdat/techniques/ms.py
```diff
@@ -331,6 +333,32 @@ def integrate_signal(self, mass, tspan, tspan_bg, ax=None):
             ax.fill_between(t, S_bg, S, color=STANDARD_COLORS[mass], alpha=0.2)
         return np.trapz(S - S_bg, t)
 
+    def integrate_flux(self, mol, tspan, tspan_bg, ax=None):
+        """Integrate a calibrated ms signal with background subtraction and evt.
+        plotting (copy of integrate_signal method)
+
+        TODO: Should this, like grab_signal does now, have the option of using a
```
ScottSoren, line 340: Pending [calculators] branch.

ScottSoren (changes requested):

AnnaWiniwarter:
Changed isotope_factor to faradaic_efficiency

PR #186, opened 2024-09-23
Enable multicomp gas calibration calculator

AnnaWiniwarter:
Updating the tutorial to work with ixdat 0.3.0, I noticed that the multicomp_gas_flux_calibration had not been updated fully to work as calculator classmethod. Fixed it and added it to NEXT_CHANGES as well. Should be an easy PR.


ScottSoren (approved):
Looks good! Thanks for catching this.

PR #196, opened 2025-08-18
Add `CalculatorPack` (grouping of calculators/calibrators into "packs")

### LINKED ISSUE #184: Saving groups of calculators
The calculators reorganization of ixdat (#182) seeks to avoid calculator redundancy - this saw the end of `ECMSCalculator` in favor of separate `MSCalculator` and `ECCalculator`s. A loss is that it was quite convenient to save an `ECMSCalculator` and use later, rather than have to separately transfer the two types of information (sensitivity factors in one, and `RE_vs_RHE`, `A_el`, and `R_Ohm` in another). A solution (see https://github.com/ixdat/ixdat/pull/182#discussion_r1750200444) could be to save multiple calculators at once, e.g. a measurements full `calculators` dictionary, in a single statement.

FrederikLizakJohansen:
This PR introduces a new CalculatorPack feature, trying to address Issue #184 and following up on the original comment by @AnnaWiniwarter in #182.

### Why it would be useful

From what I gather from the discussions, in some workflows, a measurement is analyzed using several calculators at once (e.g. an ECCalibration together with an MSCalibration). Currently, if we want to apply the same calibrations to another dataset, each calculator has to be recreated or copied separately.

`CalculatorPack` lets users capture all relevant calculators from a measurement in one step and re-apply them elsewhere, making it more **convenient** to reproduce results, while also making it more **portable** (packs are plain JSON files that can be versioned, shared, or stored with raw data), and possibly **extendable** in the future.

### Overview
`CalculatorPack` is a container for a set of calculators (EC, MS, SIQ, ...) extracted from a `Measurement`. It allows them to be:

- captured from a measurement,
- serialized to JSON or database, 
- saved/loaded from files (.ixpack.json), and
- re-attached to another measurement, with optional conflict resolution.

This makes it possible to “carry over” calibration and quantification workflows between datasets.

### Changes:

1. ixdat/calculators/packs.py _(new file)_

    - Defines `CalculatorPack` with helpers for serialization, integrity checking, file/DB I/O, and re-attachment. 
    - Skips Indexer calculators (they are always created by measurements).
    - Uses the Calculators native `as_dict()`/`from_dict()` if available. Else uses lightweight serialization.

2. ixdat/measurement_base.py _(updated)_

    - Adds integration helpers to create a pack from a measurement (`save_calculators_pack`) and apply one back (`apply_calculators_pack`).

3. development_scripts/demo_calculator_pack.py _(new demo)_

    - Demonstration of capturing EC calibrations, saving/loading packs, and re-attaching them to new measurements. **note: due to missing test data on my side, I have not been able to test multiple calculator types**.

### Still Missing / To Discuss

1. Plugin handling:

    - Some calculators (e.g. `siqCalculator`) depend on external packages (`spectro_inlets_quantification`). Currently `CalculatorPack` has no way to handle this robustly.

    - An approach might be a central plugin registry where calculator classes (or their `calculator_type`) can declare the external module and optional activation function.

    - Question: would you prefer this registry approach, or to extend each calculator with something like a `requires_plugin` attribute?

2. Broader testing:

    - Unit tests are still missing for edge cases (conflict resolution, _missing plugins_, integrity failures).
    - Only a demo script exists so far.
    
    
Closes #184

### DIFF
--- a/development_scripts/demo_calculator_pack.py
+++ b/development_scripts/demo_calculator_pack.py
@@ -1,0 +1,98 @@
+"""
+1) Read an EC measurement (Biologic .mpt)
+2) Add and apply ECCalibration (compute 'potential' and 'current')
+3) Pack calculators to file (exclude Indexer)
+4) Read pack back
+5) Read another measurement (can be same file for demo)
+6) Attach pack to target and APPLY again
+"""
+
+from __future__ import annotations
+from pathlib import Path
+
+from ixdat import Measurement
+from ixdat.calculators.ec_calculators import ECCalibration
+from ixdat.calculators.packs import CalculatorPack
+
+# --- CONFIG -------------------------------------------------------------------
+TEST_FILE = Path(__file__).parent / "../test_data/biologic/Pt_poly_cv.mpt"
+PACK_PATH = Path(__file__).parent / "demo_pack.ixpack.json"
+
+
+def compute_ec_series(meas: Measurement):
+    """Trigger calculation of EC-calibrated series to prove calibration is applied."""
+    pot = meas["potential"].data
+    cur = meas["current"].data
+    print(f"\tpotential: len={len(pot)}, min={pot.min():.3g}, max={pot.max():.3g}")
+    print(f"\tcurrent:   len={len(cur)}, min={cur.min():.3g}, max={cur.max():.3g}")
+
+
+def main():
+    # 1) Read a real EC measurement
+    print("\n[1] Reading source measurement...")
+    src = Measurement.read(TEST_FILE, reader="biologic")
+    print(
+        "\ttechnique:",
+        getattr(src, "technique", None) or getattr(src, "techniques", None),
+    )
+    print("\tseries:", [s.name for s in src.series_list][:6], "...")
+    print("\n[2] EC series before calibration...")
+    compute_ec_series(src)
+
+    # 2) Add only ECCalibration (EC-only file)
+    print("\n[3] Adding ECCalibration...")
+    ec_cal = ECCalibration(name="ec", A_el=0.196, RE_vs_RHE=0.2, R_Ohm=0.0)
+    src.add_calculator(ec_cal)
+
+    print("\tCalculators now on src:")
+    cal_list = (
+        src.calculators.values()
+        if isinstance(src.calculators, dict)
+        else src.calculators
+    )
+    for c in cal_list:
+        print("\t-", c)
+
+    # 3) APPLY EC calibration by computing series
+    print("\n[4] Computing calibrated EC series on src:")
+    compute_ec_series(src)
+
+    # 4) Build a CalculatorPack and write to file (exclude Indexer by default)
+    print("\n[5] Building CalculatorPack (excluding Indexer) and writing to file...")
+    pack = CalculatorPack.from_measurement(
+        src,
+        name="demo_pack",
+        notes="EC-only demo",
+        exclude=["Indexer", "ixdat.calculators.indexer.Indexer"],
+    )
+    print(pack.summary())
+
+    pack.to_file(str(PACK_PATH))
+    print("\tWrote:", PACK_PATH.resolve())
+
+    # 5) Read the pack back
+    print("\n[6] Reading pack back from file...")
+    pack2 = CalculatorPack.read(path=str(PACK_PATH))
+    print(pack2.summary())
+
+    # 6) Read target measurement (reuse same file for demo) and attach pack
+    print("\n[7] Reading target measurement...")
+    tgt = Measurement.read(TEST_FILE, reader="biologic")
+
+    print("\tAttaching pack to target...")
+    attached = pack2.attach_to(
+        tgt,
+        on_conflict="replace",
+    )
+    print("\tAttached calculators:")
+    for c in attached:
+        print("\t-", c)
+
+    print("\n[8] Computing calibrated EC series on target:")
+    compute_ec_series(tgt)
+
+    print("\nDone")
+
+
+if __name__ == "__main__":
+    main()
--- a/development_scripts/demo_calculator_pack.py
+++ b/development_scripts/demo_calculator_pack.py
@@ -63,7 +63,7 @@ def main():
         src,
         name="demo_pack",
         notes="EC-only demo",
-        #include = ["Indexer"],
+        # include = ["Indexer"],
     )
     print(pack.summary())
 
@@ -90,6 +90,11 @@ def main():
 
     print("\n[8] Computing calibrated EC series on target:")
     compute_ec_series(tgt)
+
+    # 7) Clean up
+    if PACK_PATH.exists():
+        PACK_PATH.unlink()
+        print(f"\n[9] Deleted temprory pack file: {PACK_PATH.name}")
 
     print("\nDone")
 
--- a/development_scripts/demo_calculator_pack.py
+++ b/development_scripts/demo_calculator_pack.py
@@ -16,7 +16,7 @@ from ixdat.calculators.packs import CalculatorPack
 
 # --- CONFIG -------------------------------------------------------------------
 TEST_FILE = Path(__file__).parent / "../test_data/biologic/Pt_poly_cv.mpt"
-PACK_PATH = Path(__file__).parent / "demo_pack.ixpack.json"
+PACK_PATH = Path(__file__).parent / "demo_pack.ixpack.yaml"
 
 
 def compute_ec_series(meas: Measurement):
--- a/development_scripts/demo_calculator_pack.py
+++ b/development_scripts/demo_calculator_pack.py
@@ -63,7 +63,7 @@ def main():
         src,
         name="demo_pack",
         notes="EC-only demo",
-        exclude=["Indexer", "ixdat.calculators.indexer.Indexer"],
+        #include = ["Indexer"],
     )
     print(pack.summary())
 
--- a/development_scripts/demo_calculator_pack.py
+++ b/development_scripts/demo_calculator_pack.py
@@ -67,7 +67,7 @@ def main():
     )
     print(pack.summary())
 
-    pack.to_file(str(PACK_PATH))
+    pack.export(str(PACK_PATH))
     print("\tWrote:", PACK_PATH.resolve())
 
     # 5) Read the pack back
--- a/src/ixdat/calculators/packs.py
+++ b/src/ixdat/calculators/packs.py
@@ -412,10 +412,11 @@ class CalculatorPack(Saveable):
         Validate the internal structure and per-calculator integrity hashes.
 
         Parameters:
-            strict: If True, treat any discrepancy as an error; otherwise aggregate warnings.
+            strict: If True, treat any discrepancy as an error;
+            otherwise aggregate warnings.
 
         Returns:
-            (ok, messages) ok=True if everything looks good (or if not strict and only warnings).
+            (ok, messages): ok=True if everything looks good
         """
         msgs: List[str] = []
         ok = True
--- a/src/ixdat/calculators/packs.py
+++ b/src/ixdat/calculators/packs.py
@@ -310,10 +310,8 @@ class CalculatorPack(Saveable):
         """
         Serialize the pack (including calculators) to a JSON-serializable dict.
 
-        Returns
-        -------
-        dict
-            The full, self-describing pack.
+        Returns:
+            dict: The full, self-describing pack.
         """
         items = self._ensure_items()
         data: JsonDict = {
@@ -333,11 +331,6 @@ class CalculatorPack(Saveable):
     def from_json(cls, data: JsonDict) -> "CalculatorPack":
         """
         Construct a pack from a JSON dictionary (inverse of `to_json`).
-
-        Notes
-        -----
-        - Calculators are *not* materialized here; only their JSON entries are stored.
-          Use `.calculators()` to materialize later.
         """
         if data.get("type") != "ixdat.CalculatorPack":
             raise ValueError(f"Unexpected type: {data.get('type')}")
@@ -356,10 +349,8 @@ class CalculatorPack(Saveable):
         """
         Write the pack to a JSON file.
 
-        Parameters
-        ----------
-        path
-            Output file path (e.g., 'my_setup.ixpack.json').
+        Parameters:
+            path: Output file path (e.g., 'my_setup.ixpack.json').
         """
         data = self.to_json()
         with open(path, "w", encoding="utf-8") as f:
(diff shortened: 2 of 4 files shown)


development_scripts/demo_calculator_pack.py
```diff
@@ -0,0 +1,98 @@
...
+    print("\n[4] Computing calibrated EC series on src:")
+    compute_ec_series(src)
+
+    # 4) Build a CalculatorPack and write to file (exclude Indexer by default)
+    print("\n[5] Building CalculatorPack (excluding Indexer) and writing to file...")
+    pack = CalculatorPack.from_measurement(
+        src,
+        name="demo_pack",
+        notes="EC-only demo",
+        exclude=["Indexer", "ixdat.calculators.indexer.Indexer"],
```
ScottSoren: I tried commenting out this line to no effect. Indexer still doesn't go into the pack.
Also, why should one have to exclude both "Indexer" and "ixdat.calculators.indexer.Indexer"? - shouldn't one be enough?

src/ixdat/calculators/packs.py
```diff
@@ -0,0 +1,559 @@
...
+    """
+    A container that serializes multiple calculators together so they can be
+    saved/loaded and re-applied to another Measurement in a single operation.
+
+    Overview:
+    - Captures a set of calculators (MS, EC, backgrounds, siq, ...)
+      with minimal metadata and per-calculator payloads.
+    - Supports:
+        `CalculatorPack.from_measurement(...)` to capture calculators
+        `pack.to_file(path)` and `CalculatorPack.read(path=...)` for file I/O
```
ScottSoren: For consistency with other ixdat methods, `to_file` should probably be either `save` (if it uses a backend) or `export` (if not)

src/ixdat/calculators/packs.py
```diff
@@ -0,0 +1,559 @@
...
+
+        # Internal storage:
+        self._calculators: List[Any] = list(calculators) if calculators else []
+        self._items: Optional[List[JsonDict]] = None  # filled by from_json or to_json
+
+        # Backing JSON for Saveable (db column). Keep it None until first to_json()
+        self.json: Optional[str] = None
+
+    @staticmethod
+    def _serialize_calculator(cal: Any) -> dict:
```
ScottSoren: Hm... could you include a sentence in the docstring about when this method is needed instead of directly using `as_dict()`?
I'm up for revisiting the whole `as_dict` serialization as part of getting ixdat a real SQL backend (e.g. the infamous #75). With that in the back of our minds, is it safe to say that having two serializations here is a temporary thing?

src/ixdat/calculators/packs.py
```diff
@@ -0,0 +1,559 @@
...
+        safe: dict[str, Any] = {"__format__": "light.dict"}
+        for k, v in vars(cal).items():
+            if k == "measurement":
+                continue  # don't persist the measurement object
+            if isinstance(v, (int, float, str, list, dict, type(None))):
+                safe[k] = v
+        return safe
+
+    @staticmethod
+    def _deserialize_calculator(cls: type, payload: dict) -> Any:
```
ScottSoren: Same question wrt why not just use `from_dict` directly, or in the long run merge these two deserialization methods.

src/ixdat/calculators/packs.py
```diff
@@ -0,0 +1,559 @@
...
+import importlib
+from datetime import datetime, timezone
+from typing import Any, Callable, Dict, Iterable, List, Optional, Sequence, Tuple, Union
+
+from ixdat.db import Saveable
+from ixdat.calculators.indexer import Indexer
+
+
+JsonDict = Dict[str, Any]
+Selector = Union[Iterable[str], Callable[[Any], bool], None]
```
ScottSoren: interesting type. Is this something that appears often in ixdat?

src/ixdat/calculators/packs.py
```diff
@@ -0,0 +1,559 @@
...
+            raise AttributeError("Measurement has no 'calculators' attribute")
+
+        # Do not include Indexes
+        calculators = [c for c in calculators if not isinstance(c, Indexer)]
+
+        # Apply include/exclude filters
+        if include is not None:
+            calculators = [c for c in calculators if _matches_selector(c, include)]
+        if exclude is not None:
+            calculators = [c for c in calculators if not _matches_selector(c, exclude)]
```
ScottSoren: I can't figure out why Indexer gets excluded even when I remove the `exclude` argument in the demo script. Can you check if that's also the case for you?

src/ixdat/calculators/packs.py
```diff
@@ -0,0 +1,559 @@
...
+            cal_ver = getattr(cal, "version", None)
+            if cal_ver:
+                item["calculator_version"] = str(cal_ver)
+
+            items.append(item)
+
+        self._items = items
+        return items
+
+    def to_json(self) -> JsonDict:
```
ScottSoren, line 261: Ooh, this looks to me like a *third* serialization method :(

src/ixdat/calculators/packs.py
```diff
@@ -0,0 +1,559 @@
...
+        pack.schema_version = str(data.get("schema_version") or cls.SCHEMA_VERSION)
+        pack.ixdat_version = data.get("ixdat_version")
+        pack.created_utc = data.get("created_utc") or _now_utc_iso()
+        pack._metadata = data.get("metadata", {})
+        pack._items = list(data.get("calculators") or [])
+        # Keep canonical db json string if caller wants to save to DB
+        pack.json = json.dumps(data, indent=2)
+        return pack
+
+    def to_file(self, path: str) -> None:
```
ScottSoren: Probably `export` to be consistent with other Saveable classes that have a way to write themselves to file that isn't through an ixdat backend.

src/ixdat/calculators/packs.py
```diff
@@ -0,0 +1,559 @@
...
+        Read a pack from a file (`path`)
+
+        Returns:
+            CalculatorPack
+        """
+        with open(path, "r", encoding="utf-8") as f:
+            data = json.load(f)
+        return cls.from_json(data)
+
+    def calculators(self) -> List[Any]:
```
ScottSoren, line 325: It can be a bit confusing that this is a method whereas `Measurement.calculators` is a dictionary. Not sure which is best practice. Would you suggest changing `Measurement.calculators` to a method to match `CalculatorPack.calculators`?

src/ixdat/calculators/packs.py
```diff
@@ -0,0 +1,559 @@
...
+
+            # Always use CalculatorPack’s deserializer
+            cal = self._deserialize_calculator(CalCls, payload)
+            calculators.append(cal)
+
+        self._calculators = calculators
+
+        return list(self._calculators)
+
+    def validate(self, *, strict: bool = True) -> Tuple[bool, List[str]]:
```
ScottSoren, line 359: Should all `Saveable` classes have a `validate` method?

src/ixdat/calculators/packs.py
```diff
@@ -0,0 +1,559 @@
...
+    ) -> List[Any]:
+        """
+        Attach all calculators in the pack to a Measurement.
+
+        Parameters:
+            measurement: Target Measurement (must provide `.add_calculator(calculator)`).
+            on_conflict:
+                "replace" (default): replace existing calculator of same class+name
+                "skip": keep existing one, skip new
+                "duplicate": add new one and allow duplicates (caller handles names)
```
ScottSoren, line 422: "duplicate" effectively will require that the calculators can be added. `Measurement.calculators` uses [https://github.com/ixdat/ixdat/blob/40b7ef213d2e013473a500c709d2714de8c6a915/src/ixdat/measurement_base.py#L525](Measurement.consolidate_calculators), which combines calculators of the same type with their `__add__` method. This can make complete sense when e.g. you have an ECCalibration with the electrode area and another ECCalibraiton with the RE potential, or MSCalibrations with sensitivity factors for different molecules. But yeah I guess then they shouldn't have the same names...
So this isn't a request for a change, just something I wanted you to know.

ScottSoren (changes requested):
Nice work!
- Why JSON instead of YAML? Yaml has the advantage that you can make comments in the files. This could be nice if I'm sending someone a text file about the nice calibrations that I've made, which they can load and apply as a CalculatorPack, but also read and see how I got the data. But there could be advantages of JSON over YAML that I don't know.
- The "Indexer" calculator gets excluded even if I don't ask for it to be, and I can't figure out why. I can't think of a case where you would need to save it in a calculator pack, but just bothers me that something doesn't behave
- Other than that, I've left some other comments in the code, with the thorniest I guess questioning the need for multiple serialization/deserialization methods.

FrederikLizakJohansen:
>     * Why JSON instead of YAML? Yaml has the advantage that you can make comments in the files. This could be nice if I'm sending someone a text file about the nice calibrations that I've made, which they can load and apply as a CalculatorPack, but also read and see how I got the data. But there could be advantages of JSON over YAML that I don't know.

Good point! Although my original intent to use JSON comes initially from the fact that (to my knowledge) `pyyaml` isn’t in our dependencies. Therefore JSON is the only format we can rely on everywhere (unless we make some changes to our dependencies ofc.) -- and even if we used YAML, comments wouldn’t survive a load/save cycle anyway. For that reason I think it’s safer to stick with JSON and encourage people to use the notes/metadata fields for any comments they want to keep. I am open to discuss this though.

development_scripts/demo_calculator_pack.py
FrederikLizakJohansen: Good catch! 
Include=["Indexer"] was acting as a filter, which left only the Indexer. I’ve changed the logic so include is additive (union) instead of restrictive. That means you keep everything else and can explicitly add Indexer back in (or anything else). Default behavior still skips Indexer when no exclude/include is given.

The "ixdat.calculators.indexer.Indexer" was just me testing out the behaviour of the "_matches_selector" helper function. It should work with either. It is now removed from the demo file.

Changes here: c8f15dfbcbde8473b3ca9c6bb4be2edefe1c4363

src/ixdat/calculators/packs.py
FrederikLizakJohansen: Resolved here: c8f15dfbcbde8473b3ca9c6bb4be2edefe1c4363

src/ixdat/calculators/packs.py
FrederikLizakJohansen: Good point.
I suggest we keep both: `save()` persists to the configured ixdat backend (IDs, table namespace, discoverable via read(id=...)), while `export()` writes a portable JSON file that can be shared.

Changes here: 3dfc03a868085d8db1cc70c60eb102e4688b6c9d

src/ixdat/calculators/packs.py
FrederikLizakJohansen: Total valid point. I must concede, I did not think to look deeper into the Saveable class before making these helper functions. It makes much more sense (and makes everything more simple) to reuse the already defined functions and rely on them being correct rather than to add extra redundancy.

I simplified the pack to use the calculators’ own `as_dict()`/`from_dict()` (inherited from Saveable) and removed the extra serializer/deserializer. The pack now just stores class_path + that payload.

For now, let's keep the `as_dict()` serialization for now. We can tackle #75 when I understand ixdat a bit better :-)

Changes here: [7c1b43d](https://github.com/ixdat/ixdat/pull/196/commits/7c1b43db167abbb38dcfc436fe9cbef271d30ec9)

src/ixdat/calculators/packs.py
FrederikLizakJohansen: See https://github.com/ixdat/ixdat/pull/196#discussion_r2352097803

src/ixdat/calculators/packs.py
FrederikLizakJohansen: I changed `Selector` → `_Selector` (made it private to `packs.py`) and expanded the `from_measurement` docstring, as well as added a comment above the type definition. Changes here: [8d98714](https://github.com/ixdat/ixdat/pull/196/commits/8d987148d8aed8af61b9c337c918ed4c49d2ab43)

The idea is just to give users flexible ways to filter calculators:
- `None` --> keep all
- list of names --> e.g. `["ECCalibration", "Indexer"]`
- predicate function --> e.g. `lambda c: isinstance(c, ECCalibration) and c.RE_vs_RHE > 0.1`

src/ixdat/calculators/packs.py
FrederikLizakJohansen: See https://github.com/ixdat/ixdat/pull/196#discussion_r2351614409

src/ixdat/calculators/packs.py
FrederikLizakJohansen: I decided to keep `to_json()`/ `from_json() `because they give us a single helper that converts a `CalculatorPack` into a plain JSON-safe dict and back again.

The point is not to introduce a second “serialization layer”, but to have one well-defined transformation between the Python object and the JSON structure we write to file. `export()` and `read()` just wrap these helpers for file I/O.

Changes here: [8ea36dd](https://github.com/ixdat/ixdat/pull/196/commits/8ea36dd550c0997b2ef299ae8a3f2df6503aef26)

FrederikLizakJohansen:
Upon further thinking, I think we can accommodate YAML files, by making pyyaml a soft optional dependency. I have made the changes in f388c00, and switched the default format in the demo to YAML.

src/ixdat/calculators/packs.py
FrederikLizakJohansen: Fair. I have changed `calculators()` to a `@property` so `pack.calculators` reads the same as `measurement.calculators`. Changes in f388c00.

src/ixdat/calculators/packs.py
FrederikLizakJohansen: I do not think all `Savable` classes need it (at least for now). `CalculatorPack` stores SHA256 hashes per calculator entry, which I believe is unique to its file-exchange use case. I've opened it as an issue in #202.

src/ixdat/calculators/packs.py
FrederikLizakJohansen: Thanks, that's a useful clarification. Worth noting in the docstring that "duplicate" bypasses `consolidate_calculators`. So if you want the `__add__`-style merging you're better off calling `measurement.add_calculator()` directly. I've added a note in the docstring in 0508dc8.

FrederikLizakJohansen:
With #209 merged, ixdat can save a measurement and its calculators together in one SQLite file, then load them again. That meets our current need to keep calibrations together, so I’m closing this PR.

I’ll leave #184 open for the remaining idea of transferring a set of calculators to another measurement.

PR #198, opened 2026-02-12
add `AsimovReader` for retrieving measurements from asimov.enci.dk

FrederikLizakJohansen:
**`AsimovReader` (src/ixdat/readers/asimov.py)**
New reader class that connects to asimov.enci.dk and retrieves measurements via REST API. Supports retrieval by `dataset_id`: Direct lookup using the asimov database ID.

The reader returns ixdat `Measurement` objects with with data fetched from the asimov API.
`KeycloakDeviceTokenProvider` (src/ixdat/auth/keycloak.py)

Authentication handler implementing the OAuth 2.0 Device Flow protocol for Keycloak. This enables passwordless authentication to asimov.enci.dk without exposing user credentials directly to ixdat. The flow:
1. Initiates authorization request
2. User visits URL and enters code
3. Exchanges authorization for access token
4. Refreshes tokens as needed

### Portable Serialization
Added `to_portable_dict()`/`from_portable_dict()` methods to:
- `Saveable` base class
- `Measurement` class
- Series classes (`ValueSeries`, `TimeSeries`, etc.)

These provide backend-agnostic serialization that doesn't reference ixdat backend classes. The current `as_dict() `method has ties to potential ixdat DB plans discussed in PR #75. These new "portable" versions provide an alternative for storing measurements in external databases and passing them over web APIs until the #75 plan is finalized.

**Note on dependency:** The asimov DB internally uses ixdat to store Measurement objects in the form of the portable dicts generated by this implementation of to_portable_dict(). Therefore, this implementation have a direct dependency on that data flow - the reader retrieves the same portable dict format that asimov stores internally, ensuring compatibility between ixdat and the asimov backend.

### Tests
- Unit test for `KeycloakDeviceTokenProvider` (tests/unit/test_keycloak_token_provider.py)
- Unit test for portable dict serialization (tests/unit/test_portable_dict.py)

### Demo
A demo script (development_scripts/reader_demonstrators/demo_asimov_reader.py) showcases:
- Retrieving measurements from asimov-db using database ID (dataset_id)
- Using Keycloak authentication with user login credentials to query for measurements


FrederikLizakJohansen:
The current commits sketch out the base implementation. Additional endpoints for asimov are scheduled, meaning more potential properties for the AsimovReader class; like a `list_datasets()` and `search_datasets()` function. Imagining:
- Filter by kind/technique (e.g., kind="ec")
- Date range filtering (created_after, created_before), names, etc.
This will require corresponding API endpoints on asimov.enci.dk (e.g., GET /api/datasets?kind=ec&limit=50).

FrederikLizakJohansen:
Added portable dict support for `Spectrum` and `SpectrumSeries` so the `AsimovReader` can handle these too, not just `Measurement`. Each portable dict now carries an `object_type` field ("measurement", "spectrum", "spectrum_series") so that the reader can automatically dispatch the correct class without needing the user to pass `cls` explicitly.

Also: Round-trip tests for `Field`, `Spectrum`, and `SpectrumSeries` added.

FrederikLizakJohansen:
After more consideration, I think it be best if we keep this PR to the base reader and portable dict, and move the search/list features to a new future PR. Created feature-issue #201 for now.

ScottSoren:
I'm having trouble testing this, possibly because I don't quite understand the demo script. Could you add some comments / documentation in that script? 
Running it in spyder gives the error 
```
demo_asimov_reader.py: error: the following arguments are required: --dataset-id
An exception has occurred, use %tb to see the full traceback.

SystemExit: 2
```
which I guess is generated by `argparse`

Finding dataset-id's on asimov is tricky, but clicking a "Measurement" from "Graph View" opens an "Inspector" which includes an "Entity ID". I copied the one from "demo_03" in project "ixdat_demo" and ran the following from a terminal (Anaconda prompt):
```
python demo_asimov_reader.py --dataset-id 3b644a7e-211c-4c24-beef-8bbc84470ee0
```
This opened a browser window asking for permissions, as described in the docstring. After I granted the permissions, a long error was printed to the Anaconda Prompt terminal, of which the important part is:
```
requests.exceptions.HTTPError: 404 Client Error: Not Found for url: https://asimov.enci.dk/api/datasets/3b644a7e-211c-4c24-beef-8bbc84470ee0
```

My normal workflow is to start with the demo script. I'll start on code review of `asimov.py` in parallel, but if you could add a bit of documentation to the demo and/or change the `if __name__ == "__main__"` section to run without arguments using some demo data as default, that would help!

FrederikLizakJohansen:
I pulled the shared portable-dict metadata handling into a couple of small helpers so we are not repeating the same `name / technique / metadata / tstamp / sample_name` boilerplate in several places, while still leaving the actual object-specific parts like `series_list`, `field`, and `durations` where they are easiest to maintain.
  
@ScottSoren I also simplified the demo flow so it no longer depends on `argparse` or on the user hunting down a valid dataset ids. The script now runs directly against a fixed pair of static example  IDs, which makes it easier to start from the demo script in the usual workflow.

development_scripts/reader_demonstrators/demo_asimov_reader.py
```diff
@@ -0,0 +1,71 @@
...
+    ax_ec_right = ax_ec.twinx()
+
+    line_styles = ("-", "--")
+    u_color = "C0"
+    j_color = "C3"
+    for idx, meas in enumerate(measurements):
+        label = _measurement_label(meas)
+        linestyle = line_styles[idx % len(line_styles)]
+        t_u, u = meas.grab(meas.U_name)
+        ax_ec.plot(
```
ScottSoren: You could use ixdat's built-in ECPlotter to get almost exactly the same thing:
```
        meas.plot(
            axes=[ax_ec, ax_ec_right],
            U_color="C0",
            J_color="C3",
            linestyle=linestyle,
        )
```
Only thing that doesn't work perfectly there is the labels for the legends.

development_scripts/reader_demonstrators/demo_asimov_reader.py
```diff
@@ -0,0 +1,71 @@
+import matplotlib.pyplot as plt
+
+from ixdat import Measurement
+
+DATASET_IDS = [
```
ScottSoren, line 5: Nice! Thanks for making this run with defaults :)

src/ixdat/auth/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .keycloak import KeycloakDeviceTokenProvider
```
ScottSoren, line 1: I wonder if Keycloak interface belongs in src/ixdat/plugins/ ? 

`plugins/__init__.py` implements a structure that avoids packages becoming hard dependencies, which may or not make sense to use for keycloak? You seem to have avoided packages (like `python-keycloak`) not in the python standard library, so that might not be necessary, so just a question of what makes most sense in code organisation. I wonder also if it might make future sense to have an `asimov` package.

Same goes, btw, for `echemdb` - that should perhaps have been a plugin, going via their [unitpackage](https://echemdb.github.io/unitpackage/) code, but that was another PR :P

src/ixdat/auth/keycloak.py
```diff
@@ -0,0 +1,249 @@
...
+                if refreshed_tokens and self._access_token_is_valid(refreshed_tokens):
+                    return refreshed_tokens["access_token"]
+
+        new_tokens = self._device_authorization_flow()
+        return new_tokens["access_token"]
+
+    def clear_cached_tokens(self):
+        """Delete local token cache if present."""
+        try:
+            self.cache_path.unlink()
```
ScottSoren, line 108: This might cause problems if we, in the future, used the "auth" cache folder for more than keycloak. "keycloak" should probably be in the name of a folder that can be deleted unilaterally by the keycloak module.

src/ixdat/tools.py
```diff
@@ -133,6 +133,31 @@ def _do_request():
             )
         raise RuntimeError(msg) from exc
 
+    except requests.exceptions.HTTPError as exc:
```
ScottSoren, line 139: Can you add some comments here? 
When would this exception arise, and what does it do?

src/ixdat/auth/keycloak.py
```diff
@@ -0,0 +1,249 @@
...
+        }
+        if self.client_secret:
+            data["client_secret"] = self.client_secret
+        try:
+            refreshed = self._post_form(self.token_endpoint, data)
+        except RuntimeError:
+            return None
+        return self._store_tokens(refreshed)
+
+    def _device_authorization_flow(self):
```
ScottSoren, line 150: This is increasingly ringing "re-inventing the wheel" bells for me (or worse, claude-rewrote-someone-elses-solution-that-you-could-have-just-used bells). Is this not already taken care of in e.g. `python-keycloak` or `python-keycloak-api` or something?
To be fair, googling things like "python-keycloak authenticate via browser" and similar isn't bringing me to it, nor am I having an easy time finding exactly that in [their documentation](https://python-keycloak.readthedocs.io/en/latest/modules/openid_client.html).

src/ixdat/auth/keycloak.py
```diff
@@ -0,0 +1,249 @@
...
+                f"  Code: {user_code}"
+            )
+        else:
+            raise RuntimeError(
+                "Keycloak device authorization response did not include a usable "
+                "verification URI."
+            )
+
+        deadline = time.time() + expires_in
+        while time.time() < deadline:
```
ScottSoren, line 204: Can you write some comments about what this loop does? 
What does it try, in what order, and when does the loop end?

src/ixdat/auth/keycloak.py
```diff
@@ -0,0 +1,249 @@
...
+        expires_in = int(details.get("expires_in", 600))
+        device_code = details.get("device_code")
+
+        if not device_code:
+            raise RuntimeError(
+                "Keycloak device authorization response did not include `device_code`."
+            )
+
+        if verification_uri_complete and self.open_browser:
+            webbrowser.open(verification_uri_complete)
```
ScottSoren: This is the only place that I can see the webbrowser is used. 
If that is missing from `python-keycloak`, and if it's the only thing that's missing, then the only thing needed in ixdat for the behaviour we want would be a single try-except in the asimov reader or an asimov plugin, right?

src/ixdat/readers/asimov.py
```diff
@@ -0,0 +1,237 @@
...
+
+        d = {**payload, "metadata": meta}
+        if not d.get("name"):
+            d["name"] = dataset.get("label") or str(dataset.get("id"))
+
+        if cls is None:
+            object_type = d.get("object_type", "measurement")
+            cls = OBJECT_TYPE_CLASSES.get(object_type, Measurement)
+
+        return cls.from_portable_dict(d, reader=self, **kwargs)
```
ScottSoren: This is, in my opinion, a wrong placement of the burden of work in a modular pluggable package.

If a new reader gets serialized data in a different format than that which exists in ixdat (even though we know the latter needs to change), then it is the job of the reader to process the data into what ixdat is built around.

A future reader could get a different type of portable dict. Should it go in and add a third parallel serialization to the entirety of ixdat? 

Let's instead write code **here** in the reader to process `d` into what `Measurement.from_dict()` works with now, like all readers do, and take the big redo of ixdat dictionaries in a seperate PR.

tests/unit/test_portable_dict.py
```diff
@@ -0,0 +1,114 @@
+import numpy as np
```
ScottSoren: Very nice to have this test!
But, as commented earlier, a new reader doesn't get to add a parallel serialization format to central ixdat classes. Save this for another PR.

src/ixdat/db.py
```diff
@@ -97,6 +97,49 @@ def get_database_name():
     return DB.backend.__class__.__name__
 
 
+def _to_portable_value(value):
```
ScottSoren: A new reader doesn't get to add a parallel serialization channel to fundamental ixdat classes. I have no doubt that it is possible to make a serialization that is superior to current ixdat dictionaries, but shifting should happen for all of ixdat, in separate PR.

src/ixdat/data_series.py
```diff
@@ -43,6 +43,53 @@ def from_dict(cls, obj_as_dict):
         series_class = SERIES_CLASSES[series_type]
         return series_class(**obj_as_dict)
 
+    @classmethod
+    def from_portable_dict(cls, dct, tseries_dict=None):
```
ScottSoren: Adding a reader should not require changes in data_series.py
Let's take this seperately. I've made some comments anyway, for that future PR.

src/ixdat/readers/asimov.py
```diff
@@ -0,0 +1,237 @@
...
+    Authentication is done via bearer token:
+
+    - Direct token: pass ``token=...`` or set ``ASIMOV_ACCESS_TOKEN``.
+    - Keycloak provider: pass ``token_provider=...``.
+    - Auto-initialise Keycloak device flow with Asimov defaults.
+      Optional env-var overrides:
+      ``KEYCLOAK_SERVER_URL``, ``KEYCLOAK_REALM``, ``KEYCLOAK_CLIENT_ID``.
+    """
+
+    def __init__(
```
ScottSoren: Can you add a docstring explaining these optional arguments that one can initialize an `AsimovReader` with?

src/ixdat/spectra.py
```diff
@@ -14,7 +14,13 @@
 
 import warnings
 import numpy as np
-from .db import Saveable, fill_object_list, PlaceHolderObject
+from .db import (
```
ScottSoren: Same thing. Something is wrong which needs to be taken care of seperately if a new reader requires changes to Spectrum.py. I don't think it does though.

src/ixdat/data_series.py
```diff
@@ -43,6 +43,53 @@ def from_dict(cls, obj_as_dict):
...
+        name = dct["name"]
+        unit_name = dct["unit_name"]
+        data = np.asarray(dct["data"])
+
+        if series_type == "tseries":
+            return TimeSeries(
+                name=name,
+                unit_name=unit_name,
+                data=data,
+                tstamp=dct.get("tstamp", 0.0),
```
ScottSoren: I have a problem with a missing tstamp passing silently. If a dataset doesn't have an absolute timestamp, ixdat might not be the right package to analyze it.

src/ixdat/data_series.py
```diff
@@ -43,6 +43,53 @@ def from_dict(cls, obj_as_dict):
...
+            if tseries_dict and "tseries_name" in dct:
+                tseries = tseries_dict.get(dct["tseries_name"])
+            return ValueSeries(
+                name=name,
+                unit_name=unit_name,
+                data=data,
+                tseries=tseries,
+            )
+        if series_type == "field":
+            axes = [cls.from_portable_dict(ax) for ax in dct.get("axes_series", [])]
```
ScottSoren: Avoid calling this `ax`, because that is what matplotlib axis objects get called.

src/ixdat/data_series.py
```diff
@@ -191,6 +238,12 @@ def data(self):
         return self._data.copy()  # TODO: make data series data immutable with numpy flag
         # see: https://github.com/ixdat/ixdat/pull/101/files#r1126172936
 
+    def to_portable_dict(self):
+        """Serialize the Field with its axes inlined."""
```
ScottSoren: Doesn't this cause duplication of data (a big no-go) when serializing a Measurement, since the axis series are also data series in the measurement? 

I can see you've prevented this from resulting in extra `tseries` in a round trip using the `tseries_dict` input to `from_portable_dict`, but that is a very fragile patch that doesn't address the duplication of data in the serializatio.

src/ixdat/data_series.py
```diff
@@ -43,6 +43,53 @@ def from_dict(cls, obj_as_dict):
...
+            return TimeSeries(
+                name=name,
+                unit_name=unit_name,
+                data=data,
+                tstamp=dct.get("tstamp", 0.0),
+            )
+        if series_type in ("vseries", "constantvalue"):
+            tseries = None
+            if tseries_dict and "tseries_name" in dct:
+                tseries = tseries_dict.get(dct["tseries_name"])
```
ScottSoren: I think this will often scramble data. There can be multiple tseries of the same name. This will be the case when appending measurements taken from the instrument, for example.

See https://github.com/ixdat/tutorials/blob/main/L3_data_structure/ixdat_demo_manipulating_electrochemistry_data.ipynb

ScottSoren (changes requested):
Hi Frederik,

Great work. However, there are serious problems in the `portable_dict` implementation, which I've commented on in data_series.py. I would much rather have the `AsimovReader` do what it needs to to get the data from asimov into a normal ixdat dictionary. I also recommend going through [this totorial](https://github.com/ixdat/tutorials/blob/main/L3_data_structure/ixdat_demo_manipulating_electrochemistry_data.ipynb) to get a better sense of what ixdat has to do (and tries to do in a smart way, though it can surely be smarter) with regards to organising its data series.

A tool that you might need instead could be an `AsimovExporter` on asimov's side, to format the data coming out of asimov. That would be a more appropriate place to put a new structure of ixdat dat, since it defeats the point of a modular structure, and would set a precedent leading to a huge proliferation of representations of ixdat data, if a PR adding a new reader could also add something like `to_portable_dict`. 

Other comments I have center around the question of whether some of the keycloak interface could be done with another package? [python-keycloak](https://github.com/marcospereirampj/python-keycloak) for example has 166 contributors, many of whom are probably professionals in exactly that - the more we can rely on something like that, the more robust I think it makes ixdat going forward. It could be that our use case is so special that it requires bottom-up coding, but then my question is why? (Some of your instinct to make a new serialization is no doubt because I kind of poorly re-invented the wheel of object-relational mapping instead of using something like SQLalchemy from the start.)

development_scripts/reader_demonstrators/demo_asimov_reader.py
FrederikLizakJohansen: My apologies for not making this in the first place ;)

src/ixdat/auth/keycloak.py
FrederikLizakJohansen: With my current (still limited) understanding of authentication flows, what I was trying to perform here with the "authenticate via browser" is a type of Device Flow; that only needs an outgoing internet connection. I think the more common browser login flow is called PKCE, but the problem with that is that it only really works on desktop apps, and breaks in things like jupyter notebooks and especially in headless environments like our VPS.

As for python-keycloak, my first instinct was actually to avoid adding any new dependencies, since I had the impression that ixdat tries to stay lean; so I tried to avoid that using just `requests` instead, which is already there. Looking at their documentation though, they do offer methods that work with Device Flow, like `device()`, `token()` and `refresh_token()` that could be used instead of some of the current custom implementations. However all they would replace is about maybe 40 lines of code, while the polling, caching and error handling would have to stay - and in exchange we then would have three maybe four new dependencies. I don't think that makes sense to do currently; but I am open to discussing this, since I agree that might be better practice to depend on maintained codebases rather than relying on custom code.

src/ixdat/auth/__init__.py
FrederikLizakJohansen: I totally agree on the `echemdb` part; I realised too late that they provided perhaps a better interface for retrieving the data. The author commented on this in #194. I will address this at some point.

Whether Keycloak belongs in the plugins depends on the decision to use python-keycloak or not. Currently addressed [here](https://github.com/ixdat/ixdat/pull/198#discussion_r3146433490), I think the easiest solution is to stick with the custom implementation, but I am up for discussing this. If we decide to depend on python-keycloak, then definitely we should consider moving it to the plugins.

src/ixdat/auth/keycloak.py
FrederikLizakJohansen: Fair point, I agree. Fixed in [1e371d6](https://github.com/ixdat/ixdat/pull/198/commits/1e371d673ab962a7205bb5cb8ba41c3c941148ea). Cache is now at `auth/keycloak/<realm>_<client_id>.json`.

development_scripts/reader_demonstrators/demo_asimov_reader.py
FrederikLizakJohansen: Good call, switched to `meas.plot()` in f86df71c390ed511c0e92e321d505e1d16bff0d0. Dropped the dataset-specific legend labels for the EC panel, since this is just a demo.

src/ixdat/tools.py
FrederikLizakJohansen: Added a comment in 75e95e7cd71f40bc971e6c2c6ee9d46831e288ef. It is triggered by HTTP responses like 4xx or 5xx, so to my knowledge things like "not found" and "server error". What this tries to do is to unravel the JSON output of the exception into a RuntimeError message instead. Just making it more clear where the error came from. For FastAPI calls say to Asimov you will get something like: "HTTP 404 Not Found: Dataset with id '3b644a7e' does not exist" if you put in an ID that does not exist. It will just fall back to raw responses if there is no JSON. I used this to debug, but I think it is just nice for the user to also get some readable text back if they make a mistake. This is also used by the `echemdb` reader, so users of this would also benefit.

src/ixdat/auth/keycloak.py
FrederikLizakJohansen: Added in 3e7575f49bad72e0a3abb9c6ff9cb2c20605b72a

src/ixdat/auth/keycloak.py
FrederikLizakJohansen: Again this depends on the decision to make it a plugin using python-keycloak (https://github.com/ixdat/ixdat/pull/198#discussion_r3146433490). The webbrowser line however is just one of multiple pieces that are handled here, like the polling and especially the caching of tokens on the disk so that the user doesn't have to log in every time - and also refreshing the tokens silently when it expires. I am trying to make the process of retrieving data from ASIMOV as frictionless as possible :-) I am up for discussing this further; I might be wrong in my approach.

FrederikLizakJohansen:
@ScottSoren; thanks for the thorough review. You were right on the big stuff. This turned out to be a bit more complex than I first thought, so bear with me here; I am just gonna one reply here since the remaining review comments fit well together. The changes made here are in 4464fb7ec6e727d7ff6a6566a9a4ca54be34de18.

**The big one: `portable_dict` in core**
You're right. A reader shouldn't reshape ixdat's core just because its format looks different. I simply hadn't thought this through. So I pulled it out again:
- `db.py`, `data_series.py`, `measurement_base.py`, `spectra.py`: every `to_portable_dict` / `from_portable_dict` is gone. Those files are back to pre-PR state.
- `tests/unit/test_portable_dict.py` deleted.
- The payload-to-ixdat translation now lives in the reader: `AsimovReader._build_kwargs` and `AsimovReader._build_series`. I've tried to keep this as simple as can be, but please let me know if you have other suggestions. All importing on the asimov-server-side is handled exclusively by Asimov now to not bloat this reader with functionality that it does not own.
- New payload-parsing tests in `tests/unit/test_asimov_reader.py`.

The "ixdat dicts should be redone" conversation stays a separate PR, as you suggested.

**Fixing the name-collision and the duplication of data**
The old payload pointed at a TimeSeries from a ValueSeries using its `name` string, and a Field's t-axis was inlined under `axes_series` even when the same TimeSeries was already in `series_list`. Both were really broken.

To fix the naming-collision I had to redo the payload format itself on the server-side. Asimov's writer and the reader now both use payload-local keys. Every entry in `series_list` carries a key like `"s0"`, `"s1"`, a ValueSeries points at its TimeSeries with `tseries_key`, and a Field whose axis is also a top-level series uses `axes_keys: ["s0"]` instead of inlining. I went with short string keys (`"s0"`) over plain integers mostly because they read as "series 0" in a JSON, and they don't get mixed up with backend ids from the ixdat backend (more on this below). The Asimov writer was updated alongside this, and all stored payloads were re-parsed into the new shape, so there should be no back-compat issues.

While going through this, I realized that I might be re-inventing the wheel once again, and so I went into a rabbit hole of trying to figure out if the reader could just lean on ixdat's `as_dict()`/`from_dict()` directly and skip the custom translation; since ofc the `key`/`tseries_key`/`axes_keys` pattern is doing more or less the same job as ixdat's existing `s_ids`/`a_ids`. I tried two versions. One where Asimov serializes with `as_dict()` and the reader calls `from_dict()` directly, and one with a small `AsimovExport` along the lines of what you proposed, so that it could go through ixdat's backend hooks. Both "solutions" came out approx the same size as the current implementation, but ofc they were tied closely to ixdat's `column_attrs`/`extra_linkers` setup, which is the part being redone in #75, and I also ran into the `SpectrumSeries.durations`/`continuous` thing that `as_dict` doesn't pick up. So I dropped the experimentation and stuck with the explicit `s0/s1` shape here. It's the same id-based reference idea, just in a format Asimov and the reader can own, so nothing has to move when #75 lands. When that PR happens, this will probably end up looking a lot like whatever the new format becomes, and the custom bits on both sides can go away. But I would rather that this implementation work before we delve into #75, which this PR would then otherwise depend heavily on.

I hope you can see the point in doing this; else let me know!

**New: AsimovConfig**
I also pulled all the connection vars out of `AsimovReader.__init__` and into a small `AsimovConfig` dataclass. Before, things like base_url, Keycloak realm, client ID, timeouts, and cache path were either kwargs on the reader or scattered constants. That made the reader signature long and messy. Now:
- `AsimovConfig` holds connection + auth settings, with one big docstring explaining each field.
- `AsimovConfig.from_env()` layers env vars on defaults (`ASIMOV_BASE_URL`, `KEYCLOAK_REALM`, etc.) so CI and headless runs just set env vars.
- A module-level `ASIMOV_CONFIG` is built at import. Power users can mutate it before the first read, or pass their own `AsimovConfig` for tests. The idea is "you don't need to touch any of this unless you're on a private deployment or in a headless box."
- Added full docstrings on `AsimovConfig` and `AsimovReader.__init__`, covering the fields, env vars, and when you'd touch them.

**Already pushed earlier in the branch**
- Cache path now `auth/keycloak/<realm>_<client_id>.json`: 1e371d6
- Comment on the HTTP-error JSON unraveling: 75e95e7
- Comments on the device-flow polling loop: 3e7575f
- Demo uses `meas.plot()`: f86df71

**Still open between us**
The python-keycloak / plugins / webbrowser thread: I haven't touched it yet, since I think we should decide together. My opinion is still that device flow + token cache + silent refresh is enough custom code that python-keycloak only replaces ~40 lines, in trade for 3–4 new deps. But I'm open. If we go that way, moving the whole module under `plugins/` makes sense at the same time. Your call.

**Other**
While testing I hit a `ResourceWarning` from `webbrowser.open`. Replaced it with a small `_open_in_browser` helper in `keycloak.py`.

src/ixdat/data_series.py
FrederikLizakJohansen: Agreed, addressed in 4464fb7ec6e727d7ff6a6566a9a4ca54be34de18. `0.0` is worse than `None` as well because it looks like it was from 1970 :D. Measurements must have an absolute `tstamp` (to align `TimeSeries`), but Spectrum/SpectrumSeries can have `tstamp=None` it looks like.

src/ixdat/auth/keycloak.py
ScottSoren: I'm convinced now! Keep the device flow code here :) 
Thanks for looking into it.

src/ixdat/data_series.py
FrederikLizakJohansen: Addressed in https://github.com/ixdat/ixdat/pull/198#issuecomment-4329890619

src/ixdat/data_series.py
FrederikLizakJohansen: Adressed in https://github.com/ixdat/ixdat/pull/198#issuecomment-4329890619

src/ixdat/data_series.py
FrederikLizakJohansen: Agreed, I will try to remember it in the future! Should be addressed in https://github.com/ixdat/ixdat/pull/198#issuecomment-4329890619

src/ixdat/spectra.py
FrederikLizakJohansen: Addressed in https://github.com/ixdat/ixdat/pull/198#issuecomment-4329890619

src/ixdat/readers/asimov.py
FrederikLizakJohansen: Totally agree now that I've had a second look; addressed in https://github.com/ixdat/ixdat/pull/198#issuecomment-4329890619.

tests/unit/test_portable_dict.py
FrederikLizakJohansen: Addressed in https://github.com/ixdat/ixdat/pull/198#issuecomment-4329890619

src/ixdat/db.py
FrederikLizakJohansen: Addressed in https://github.com/ixdat/ixdat/pull/198#issuecomment-4329890619

src/ixdat/data_series.py
FrederikLizakJohansen: Addressed in https://github.com/ixdat/ixdat/pull/198#issuecomment-4329890619

src/ixdat/readers/asimov.py
FrederikLizakJohansen: Addressed in https://github.com/ixdat/ixdat/pull/198#issuecomment-4329890619

src/ixdat/auth/keycloak.py
ScottSoren: Thanks! But what does "poll the token endpoint" mean?

tests/functional/test_backends.py
```diff
@@ -1,4 +1,4 @@
-""""Tests that an ECMeasurement read from test data behaves as it should"""
+""" "Tests that an ECMeasurement read from test data behaves as it should"""
```
ScottSoren: still looks like an extra "?

src/ixdat/tools.py
ScottSoren: Just need a couple more words, for context for the "layman". ixdat code should ideally be readable to scientists who code for data analysis, even if they aren't computer scientists. Is this catching non-existant url's, missing internet connections, login/access problems, all of the above or something else?

src/ixdat/readers/asimov.py
```diff
@@ -0,0 +1,387 @@
...
+from ..tools import request_with_retries
+
+OBJECT_TYPE_CLASSES = {
+    "measurement": Measurement,
+    "spectrum": Spectrum,
+    "spectrum_series": SpectrumSeries,
+}
+
+
+@dataclass
```
ScottSoren, line 23: dataclass is new to me :) thanks! I've used `attrs` before for the same. For curiosity: https://www.attrs.org/en/stable/why.html
It seems like this means that we're killing python 3.6 compatability, though. I think that's fine - we already had to stop testing for it in the continuous integration. We'll have to warn about that somehow when releasing this on PyPI. There are a few groups around the world using ixdat for EC-MS who might be affected. Also, Spectro Inlets, who have built some things on top of ixdat.

Let's sit together when releasing this and work through it.

src/ixdat/readers/asimov.py
```diff
@@ -0,0 +1,387 @@
...
+    keycloak_client_secret: "str | None" = None
+    keycloak_scope: str = "openid offline_access"
+    keycloak_cache_path: "str | None" = None
+    keycloak_open_browser: bool = True
+    connect_timeout: float = 3.0
+    read_timeout: float = 10.0
+    total_timeout: float = 30.0
+
+    # Mapping of field name -> environment variable name.
+    _ENV_VARS = {
```
ScottSoren, line 85: I'm wondering now if `AsimovConfig` should inherit from a `KeyCloakConfig` and just overwrite `_ENV_VARS`?

src/ixdat/readers/asimov.py
```diff
@@ -0,0 +1,387 @@
...
+
+# Module-level config, which is populated from environment variables at import.
+# Edit attributes on this object before the first reader call to override.
+ASIMOV_CONFIG = AsimovConfig.from_env()
+
+
+class AsimovReader:
+    """Read a remote dataset version from Asimov into an ixdat object.
+
+    Typical use needs no arguments at all; just ``Measurement.read(<id>,
```
ScottSoren: a bit of a self-contradiction here. Better: "Typical use only needs the dataset-id as argument; ..."

Btw, the usage of AsimovReader is really nice :D

src/ixdat/readers/asimov.py
```diff
@@ -0,0 +1,387 @@
...
+        self.connect_timeout = self.config.connect_timeout
+        self.read_timeout = self.config.read_timeout
+        self.total_timeout = self.config.total_timeout
+
+        self._token = token or os.environ.get("ASIMOV_ACCESS_TOKEN")
+        self.token_provider = token_provider
+
+        if self._token is None and self.token_provider is None:
+            self.token_provider = KeycloakDeviceTokenProvider(
+                server_url=self.config.keycloak_server_url.rstrip("/"),
```
ScottSoren: Shouldn't the `rstrip` be in the `__init__` of `KeyCloakDeviceTokenProvider` or `AsimovConfig` instead? Just feel like this is a non-intuitive place to find it, should it (against expectations) be involved in a future bug.

src/ixdat/readers/asimov.py
```diff
@@ -0,0 +1,387 @@
...
+                    series_list.append(ts)
+            for s in dct["series_list"]:
+                if s.get("series_type") != "tseries":
+                    built = self._build_series(s, key_map)
+                    key_map[s["key"]] = built
+                    series_list.append(built)
+            obj["series_list"] = series_list
+            obj["aliases"] = dct.get("aliases") or {}
+
+        if "field" in dct:
```
ScottSoren, line 280: Can you add a comment here reminding the code reader when "field" will be in dct?
`# this will be the case when retreiving a Spectrum, SpectrumSeries or SpectroMeasurement`

src/ixdat/readers/asimov.py
```diff
@@ -0,0 +1,387 @@
...
+                obj["durations"] = dct.get("durations")
+                obj["continuous"] = dct.get("continuous", False)
+
+        return obj
+
+    @staticmethod
+    def _build_series(dct, key_map=None):
+        """Build a DataSeries from one payload entry.
+
+        key_map resolves tseries_key / axes_keys references when
```
ScottSoren, line 293: nice!

src/ixdat/readers/asimov.py
```diff
@@ -0,0 +1,387 @@
...
+            if key_map and "tseries_key" in dct:
+                ts = key_map.get(dct["tseries_key"])
+            return ValueSeries(name=name, unit_name=unit_name, data=data, tseries=ts)
+        if kind == "field":
+            # axes_keys references top-level series; axes_series is the inline
+            # form used by standalone Spectrum / SpectrumSeries payloads.
+            if "axes_keys" in dct and key_map:
+                axes = [key_map[k] for k in dct["axes_keys"]]
+            else:
+                axes = [
```
ScottSoren, line 324: when would you ever get to this `else`?

src/ixdat/readers/asimov.py
```diff
@@ -0,0 +1,387 @@
...
+                    AsimovReader._build_series(a, key_map)
+                    for a in dct.get("axes_series", [])
+                ]
+            return Field(name=name, unit_name=unit_name, data=data, axes_series=axes)
+
+        return DataSeries(name=name, unit_name=unit_name, data=data)
+
+    def _select_dataset_version(self, versions, version=None, version_id=None):
```
ScottSoren, line 332: Can I get you to write docstrings for all these helper methods, even though they aren't in the API? 
The docstring here could explain, for example, when will a dataset have multiple versions?

development_scripts/reader_demonstrators/demo_asimov_reader.py
```diff
@@ -0,0 +1,50 @@
...
+        linestyle = line_styles[idx % len(line_styles)]
+        meas.plot(
+            axes=[ax_ec, ax_ec_right],
+            U_color="C0",
+            J_color="C3",
+            linestyle=linestyle,
+        )
+
+    for idx, meas in enumerate(measurements):
+        meas.as_cv().plot(
```
ScottSoren: `meas.plot_vs_potential()` does the same.

src/ixdat/readers/asimov.py
```diff
@@ -0,0 +1,387 @@
...
+                f"Dataset version {dataset_version.get('id')} has no "
+                "payload_json or payload_uri."
+            )
+
+        self.dataset = dataset
+        self.dataset_version = dataset_version
+
+        # Inject Asimov provenance (dataset versioning) into metadata
+        meta = dict(payload.get("metadata") or {})
+        meta["asimov"] = {
```
ScottSoren, line 216: All these metadata fields have to do with the retreiving of the measurement rather than the measurement itself - perhaps they should instead be stored in the instance of the reader. A measurement can have a reference to the reader that created it, if the reader uses:
`return cls.from_dict(obj_as_dict, ..., reader=self)`
Could potentially be usefull if some downstream code wanted to check something in the particular asimov config of this dataset, enabling the following:
```
from ixdat import Measurement

meas = Measurement.read(... reader="asimov")
print(meas.reader.config.base_url)
```

A weakness, though, is that the reader doesn't get saved by `meas.save()`, whereas metadata attribute does get saved.

src/ixdat/readers/asimov.py
```diff
@@ -0,0 +1,387 @@
...
+            for v in versions:
+                if v.get("id") == version_id:
+                    return v
+            raise ValueError(f"No dataset version with id={version_id}.")
+        if version is not None:
+            for v in versions:
+                if v.get("version") == version:
+                    return v
+            raise ValueError(f"No dataset version with version={version}.")
+        return sorted(versions, key=lambda v: v.get("created_at", ""), reverse=True)[0]
```
ScottSoren, line 362: This requires some explanation in a comment.

src/ixdat/readers/asimov.py
```diff
@@ -0,0 +1,387 @@
...
+            read_timeout=self.read_timeout,
+            total_timeout=self.total_timeout,
+            retries=2,
+        )
+        return response.json()
+
+    def _get_json(self, endpoint, headers, params=None):
+        url = urljoin(self.base_url + "/", endpoint.lstrip("/"))
+        try:
+            response = request_with_retries(
```
ScottSoren, line 416: Can you add a comment here? What type is response?

src/ixdat/readers/asimov.py
```diff
@@ -0,0 +1,387 @@
...
+            obj["sample_name"] = dct["sample_name"]
+
+        if "series_list" in dct:
+            # Build TimeSeries first so vseries / fields can reference them.
+            key_map = {}
+            series_list = []
+            for s in dct["series_list"]:
+                if s.get("series_type") == "tseries":
+                    ts = self._build_series(s)
+                    key_map[s["key"]] = ts
```
ScottSoren, line 268: Where does `s["key"]` get assigned?
If it's within asimov, can you explain that in a comment?

tests/unit/test_asimov_reader.py
```diff
@@ -0,0 +1,270 @@
...
+            },
+        ],
+    }
+
+    obj = _reader()._build_kwargs(payload)
+    tseriess = [s for s in obj["series_list"] if isinstance(s, TimeSeries)]
+    vseriess = [s for s in obj["series_list"] if isinstance(s, ValueSeries)]
+    t0, t1 = tseriess
+    v0, v1 = vseriess
+    assert t0 is not t1  # distinct objects despite shared name
```
ScottSoren, line 231: Does the following work? 
```
np.testing.assert_allclose(obj["time/s"].data, [0, 1, 100, 101])
```

tests/unit/test_asimov_reader.py
```diff
@@ -0,0 +1,270 @@
...
+            },
+        ],
+    }
+
+    obj = _reader()._build_kwargs(payload)
+    top_level_t = next(s for s in obj["series_list"] if isinstance(s, TimeSeries))
+    field = next(
+        s for s in obj["series_list"] if not isinstance(s, (TimeSeries, ValueSeries))
+    )
+    assert field.axes_series[0] is top_level_t
```
ScottSoren, line 276: nice.

ScottSoren (changes requested):
Very nice! Thanks for the thoughtful addressing of my previous review. Mainly questions remaining, plus some small change requests, mostly asking for comments.
I like your solution with "key", and though I also in principle think that it should be covered by the id's of ixdat objects, I know that the latter are a bit of a mess at present.
I'd like to see where the "key"s get set, though, for my curiosity and thinking ahead to #75 ! Can you point me to it in the asimov code base?

tests/functional/test_backends.py
FrederikLizakJohansen: This is outdated; I removed the edit entirely in the latest commit to not mess with files not involved in this reader. I accidentally black-formatted the entire tests/ folder and it made some changes. This mistake here should be covered in another (cleanup) PR perhaps :-)

src/ixdat/readers/asimov.py
FrederikLizakJohansen: Ah okay haha, this is where it shows when we both entered the python programming scene! I really only started programming for real in python when 3.7 was released, so I have only used that. We should definitely think about how to release this in a way that does not break current users; but I also think that we could argue that they should be using >=3.7 in 2026 ;-) 

Lets agree fully on this before we merge.

src/ixdat/tools.py
FrederikLizakJohansen: I have given it another shot. While we are at it, I have added some additional comments throughout tools.py in de0c6813a01da79da9e8e2bea69ec6fe349a6f12, to make each part perhaps a little more understandable.

src/ixdat/auth/keycloak.py
FrederikLizakJohansen: I have added some more "layman" comments throughout keycloak.py in 00cdae27bd49790853acd935b39292cec142ea2d to hopefully make sense of this. Polling here basically means repeatingly asking the server if the exchange has succeeded: "has the user logged in yet".

src/ixdat/readers/asimov.py
FrederikLizakJohansen: Considered it, but I only think that would make sense in the long run end up using Keycloak settings in multiple places. I think inheritance would be a bit more structure than we need right now. AsimovConfig also holds Asimov-specific API and timeout settings.

src/ixdat/readers/asimov.py
FrederikLizakJohansen: As far as I understand `reader=self` is already being passed and stored on the measurement, so `meas.reader.config.base_url` works already. I'd like to keep the meta["asimov"] block, so that fields like `dataset_id` and `dataset_version` are the persistent answer to "where did this data come from?", and that answer needs to survive meas.save() + reload.

I tested this:
```
>>> from ixdat import Measurement
importing ixdat v0.3.1.dev1 from /home/frederik/postdoc/dev/ixdat/src/ixdat/__init__.py
>>> meas = Measurement.read("8e935aa3-adcf-4a05-a204-25759e63ca0b", reader='asimov')
>>> meas.reader
<ixdat.readers.asimov.AsimovReader object at 0x7f1b7eae7380>
>>> meas.reader.config
AsimovConfig(base_url='https://asimov.enci.dk/api', keycloak_server_url='https://auth.enci.dk', keycloak_realm='master', keycloak_client_id='ixdat-cli', keycloak_client_secret=None, keycloak_scope='openid offline_access', keycloak_cache_path=None, keycloak_open_browser=True, connect_timeout=3.0, read_timeout=10.0, total_timeout=30.0)
>>> meas.reader.config.base_url
'https://asimov.enci.dk/api'
```

tests/unit/test_asimov_reader.py
FrederikLizakJohansen: Yes! 913d1b7dc8bc295e90e3e93caf8ceb2a753e4110

src/ixdat/readers/asimov.py
FrederikLizakJohansen: Tried to explain it here 913d1b7dc8bc295e90e3e93caf8ceb2a753e4110

src/ixdat/readers/asimov.py
FrederikLizakJohansen: Good catch, tried to explain here 913d1b7dc8bc295e90e3e93caf8ceb2a753e4110

src/ixdat/readers/asimov.py
FrederikLizakJohansen: I agree, did that here: 913d1b7dc8bc295e90e3e93caf8ceb2a753e4110

src/ixdat/readers/asimov.py
FrederikLizakJohansen: Good catch! `key_map` was actually never passed to `_build_series` for the field case, even when it existed (e.g. for a `SpectroMeasurement` with both `series_list` and `field`). Fixed it by initialising `key_map = None`. I also added a comment on the `else` branch explaining it. When reading a standalone `Spectrum` or `SpectrumSeries` there is no shared `series_list` and therefore no `key_map`. See 913d1b7dc8bc295e90e3e93caf8ceb2a753e4110

FrederikLizakJohansen: Good catch! `key_map` was actually never passed to `_build_series` for the field case, even when it existed (e.g. for a `SpectroMeasurement` with both `series_list` and `field`). Fixed it by initialising `key_map = None`. I also added a comment on the `else` branch explaining it. When reading a standalone `Spectrum` or `SpectrumSeries` there is no shared `series_list` and therefore no `key_map`. See 913d1b7dc8bc295e90e3e93caf8ceb2a753e4110

src/ixdat/readers/asimov.py
FrederikLizakJohansen: Good idea, tried to do it here: 913d1b7dc8bc295e90e3e93caf8ceb2a753e4110

src/ixdat/readers/asimov.py
FrederikLizakJohansen: Yes, here 913d1b7dc8bc295e90e3e93caf8ceb2a753e4110

src/ixdat/readers/asimov.py
FrederikLizakJohansen: You are right to question this; I think it belongs in an __post_init__, so I moved it there instead, see 913d1b7dc8bc295e90e3e93caf8ceb2a753e4110

src/ixdat/readers/asimov.py
FrederikLizakJohansen: Haha, good catch, changed in 913d1b7dc8bc295e90e3e93caf8ceb2a753e4110

development_scripts/reader_demonstrators/demo_asimov_reader.py
FrederikLizakJohansen: Right! See 913d1b7dc8bc295e90e3e93caf8ceb2a753e4110

src/ixdat/readers/asimov.py
FrederikLizakJohansen: We need a `__post_init__` since `__init__` is autogenerated in the `@dataclass`, so we cannot put any logic in there.

FrederikLizakJohansen (commented):
I do not know why I need to "submit review" here since I just commented on your review. But here it is.

src/ixdat/readers/asimov.py
ScottSoren: okay! I see that's a bit tedious, but I guess easier for the future developer to find? Happy either way.

ScottSoren:
All comments resolved except for the one reminding us to deal with dropping python 3.6 support when releasing.
Great work! Approved and merging :D
Can you point me to where in the asimov code base `key_map` gets built? Might be that that's the way ixdat should do it in the future.

FrederikLizakJohansen:
@ScottSoren Thanks for the review and the merge!
 
In asimov, the key map is built in `services/api/parsing/ixdat_payload.py:172`:

```python
all_series = list(getattr(measurement, "series_list", []) or [])
series_key_map: dict[int, str] = {id(s): f"s{i}" for i, s in enumerate(all_series)}
```
Here `series_list` is ixdat’s list of data-series objects attached to the measurement

PR #200, opened 2026-04-11
Bruker NMR Reader

FrederikLizakJohansen:
# Add Bruker TopSpin NMR reader

## What this PR does

Adds a reader for **Bruker TopSpin** NMR experiment folders so that solution-NMR data can be loaded into ixdat.

```python
from ixdat import Spectrum

spec = Spectrum.read("path/to/topspin_experiment_folder", reader="bruker")

spec.plot()                    # intensity vs. chemical shift (ppm)
print(spec.metadata["NS"])        # number of scans
print(spec.metadata["SOLVENT"])   # e.g. "Urine"
```

`spec` is an `NMRSpectrum` (a new technique class added in this PR) and behaves like any other ixdat `Spectrum`.

## The "TopSpin experiment folder":

The folder produced by Bruker's TopSpin NMR software. Typically looks like this:

```
my_experiment/
├── acqus            # acquisition parameters (text, JCAMP format)
├── fid              # raw time-domain data (binary)
├── pulseprogram
└── pdata/
    └── 1/
        ├── procs    # processing parameters (text)
        ├── 1r       # processed real spectrum (binary)
        └── 1i       # processed imaginary spectrum (binary)
```

You point the reader at the top folder (`my_experiment/`) and it figures out the rest.

## How it works

The reader uses [[nmrglue](https://nmrglue.readthedocs.io/)](https://nmrglue.readthedocs.io/) (Python package for NMR file I/O) to do the actual binary parsing. ixdat then wraps the result into its standard `Spectrum` data model:

- **Processed data:** reads `pdata/<procno>/`, returns the real spectrum with a proper chemical-shift axis in ppm.
- **FID fallback:** if no processed data is available (or you pass `prefer_processed=False`), the reader returns the magnitude of the raw FID against an index axis. Phasing/Fourier transform/referencing is left to the user.

`nmrglue` is **not** added to ixdat's required dependencies, it's only imported lazily inside `BrukerNMRReader.read()`. If a user tries to read a Bruker folder without it installed, they get a clear `ReadError` explaining how to install it (`pip install nmrglue`).

## Metadata

All the key acquisition parameters from `acqus` are moved onto `spec.metadata` with their original Bruker names (`PULPROG`, `SOLVENT`, `NS`, `DS`, `TE`, `BF1`, `SFO1`, `O1`, `SW`, `TD`, `NUC1`, `DATE`, ...), and processing parameters from `procs` are exposed under `proc_*` prefixes (`proc_SI`, `proc_SF`, `proc_OFFSET`, `proc_LB`, `proc_PHC0`, ...). The full `acqus` and `procs` dictionaries are also preserved under `spec.metadata["acqus"]` / `spec.metadata["procs"]` so nothing is lost. Everything in `metadata` is JSON-serialisable.

The acquisition `DATE` is parsed into ixdat's standard unix timestamp (`spec.tstamp`) so NMR data can be co-plotted on a shared time axis with other techniques.

## New technique class

Adds `NMRSpectrum` and `NMRSpectrumSeries` in `ixdat.techniques.nmr`, following the same pattern as the existing `FTIRSpectrum` / `FTIRSpectrumSeries`. Registered as `"NMR"` and `"NMR_spectra"` in `TECHNIQUE_CLASSES`.

## Tests

A new `tests/functional/test_bruker_reader.py` is added.

The tests run against a real TopSpin 1D ¹H NMR dataset vendored under `test_data/bruker/MTBLS1_ADG19007u_162_10/`. It comes from MetaboLights study MTBLS1 (a publicly availiable human-urine 1H NMR dataset, 700 MHz, noesypr1d, 128 scans), MIT-licensed reference repository. I included full attribution in `test_data/bruker/MTBLS1_ADG19007u_162_10/CREDITS.txt`. Only the files strictly needed by the tests are vendored (~770 KB total).

## Demo script

`development_scripts/reader_demonstrators/demo_bruker_reader.py` :

1. Reads the test-data via `Spectrum.read(..., reader="bruker")`.
2. Prints the basics (class, technique, name, tstamp, ppm range, point count, intensity range).
3. Prints the lifted acquisition parameters (`PULPROG`, `SOLVENT`, `NUC1`, `BF1`, `NS`, `TE`, ...) and processing parameters (`SI`, `SF`, `OFFSET`, `LB`, `PHC0`, ...).
4. Plots intensity vs. chemical shift with the conventional inverted ppm axis, titled with the nucleus / pulse program / field / number-of-scans / solvent.

## Files added / changed

**New:**
- `src/ixdat/readers/bruker.py`:  `BrukerNMRReader` implementation.
- `src/ixdat/techniques/nmr.py`:  `NMRSpectrum` / `NMRSpectrumSeries`.
- `tests/functional/test_bruker_reader.py`
- `development_scripts/reader_demonstrators/demo_bruker_reader.py`: demo
- `test_data/bruker/MTBLS1_ADG19007u_162_10/`: test dataset

**Modified:**
- `src/ixdat/readers/__init__.py`:  registers `"bruker"` in `SPECTRUM_READER_CLASSES`.
- `src/ixdat/techniques/__init__.py`:  registers `"NMR"` and `"NMR_spectra"` in `TECHNIQUE_CLASSES`.
- `README.rst`:  adds an "NMR" row to the techniques/readers table. (EDIT: Accidentally removed the nordic-reader line and then added it back by mistake on main-branch; so we have patched it back here from main to not have merge-issues)
- `NEXT_CHANGES.rst`: entries describing the new reader and the new technique classes.

## Things out of scope for now:

- **2D and higher-dimensional NMR experiments.** This reader is scoped to 1D experiments for the time being. Support for 2D could be a future addition.
- **FT/phasing/referencing of raw FIDs.** The FID fallback intentionally doesn't try to do this; users who need that should preprocess the data in TopSpin (or with nmrglue) and let ixdat read the processed spectrum. We could eventually incorporate this as something like a new `Calculator`


FrederikLizakJohansen:
Addressing [this discussion](https://github.com/ixdat/ixdat/pull/198#discussion_r3141576024) preemptively since this reader also depends on an external package.

I think the lazy import at call-time is the right approach here rather than going through `plugins/`. The plugins system seems designed for packages that change how ixdat behaves globally. `siq` rewires mass spec quantification, `cinfdata` sets up a database connection. `nmrglue` is just a file parser that we need for one specific `read()` call. Having the user call something like `plugins.activate_nmrglue()` before reading a file seems like unnecessary setup.

src/ixdat/readers/bruker.py
```diff
@@ -0,0 +1,279 @@
...
+#   TD       - number of raw data points in the FID; for quadrature acquisition
+#              (the standard) TD is twice the number of complex points, because
+#              real and imaginary channels are stored interleaved
+#   PULPROG  - pulse program name (defines the RF pulse sequence, e.g. zg, noesypr1d)
+#   AQ_mod   - acquisition mode (e.g. DQD = digital quadrature detection)
+#   NUC1     - nucleus assigned to frequency channel 1 (e.g. 1H, 13C, 31P)
+#   P        - array of rectangular pulse lengths in microseconds (P[1] is typically
+#              the 90-degree hard pulse that tips magnetisation from z to the xy-plane)
+#   D        - array of inter-pulse delay times in seconds
+_ACQUS_KEYS = (
```
ScottSoren: Why hide this variable?
If it was public, then the demo script could import it instead of repeating it: 
```
from ixdat.readers.bruker import ACQUS_KEYS as acq_keys
from ixdat.readers.bruker import PROCS_KEYS as proc_keys
```

development_scripts/reader_demonstrators/demo_bruker_reader.py
```diff
@@ -0,0 +1,98 @@
...
+print(
+    f"x ({spec.xseries.unit_name}) : "
+    f"{spec.x.min():.2f} .. {spec.x.max():.2f}  ({spec.x.size} points)"
+)
+print(f"y           : {spec.y.min():.3g} .. {spec.y.max():.3g}")
+print()
+
+# 3. Print the lifted acquisition + processing parameters.
+md = spec.metadata
+acq_keys = (
```
ScottSoren: You could import these from the reader module instead, to show the user of the demo script that they're available for import there :)

src/ixdat/readers/bruker.py
```diff
@@ -0,0 +1,279 @@
...
+
+
+# Acquisition parameters lifted from `acqus` into a flat metadata dict.
+# Anything not in this list is still preserved in metadata["acqus"].
+#
+# Definitions follow the Bruker TopSpin Acquisition Commands and Parameters
+# Reference Manual. A copy is available at:
+# https://www.nmr.ucdavis.edu/sites/g/files/dgvnsk4156/files/inline-files/acqu_commands_parameters.pdf  # noqa: E501
+#
+# Key Bruker acqus abbreviations:
```
ScottSoren, line 19: Very nice to have this in the comments! :)

test_data/bruker/MTBLS1_ADG19007u_162_10/pdata/1/proc
```diff
@@ -0,0 +1,99 @@
...
+##$STSI= 0
+##$STSR= 0
+##$SW_p= 14005.6022408964
+##$SYMM= 0
+##$S_DEV= 0
+##$TDeff= 0
+##$TDoff= 0
+##$TI= <Study Number ADG19007
+Compound PPAR pan
+Human Urine
```
ScottSoren, line 83: ?

src/ixdat/readers/bruker.py
```diff
@@ -0,0 +1,279 @@
...
+            path_to_file (Path or str): Path to the experiment folder (the one
+                containing ``acqus``). May also be the path to a file inside
+                that folder, in which case the parent folder is used.
+            name (str): Optional name override. Defaults to the folder name.
+            cls (Spectrum subclass): Class to instantiate. Defaults to
+                :class:`NMRSpectrum`. Anything that is not a subclass falls
+                back to :class:`NMRSpectrum`.
+            procno (int): Which processed-data subfolder to read from
+                (``pdata/<procno>``). Defaults to 1.
+            prefer_processed (bool): If True (default) and a processed
```
ScottSoren: And what if I want both? I call it once with prefer_processed=True and one with prefer_processed=False?

development_scripts/reader_demonstrators/demo_bruker_reader.py
```diff
@@ -0,0 +1,98 @@
...
+    if k in md:
+        print(f"  {k:14s} = {md[k]}")
+print()
+print(
+    f"full acqus dict has {len(md['acqus'])} keys; "
+    f"full procs dict has {len(md['procs'])} keys."
+)
+
+# 4. Plot the spectrum. Chemical-shift convention: high ppm on the left.
+fig, ax = plt.subplots()
```
ScottSoren: You could consider making an `NMRPlotter` which inverts the xaxis, so that `spec.plot()` outputs exactly what NMR people like to look at.

src/ixdat/readers/bruker.py
```diff
@@ -0,0 +1,279 @@
...
+        if not (folder / "acqus").is_file():
+            raise ReadError(
+                f"No 'acqus' file found in {folder}. Point BrukerNMRReader at a "
+                "Bruker TopSpin experiment folder (the one containing acqus)."
+            )
+
+        self.path_to_folder = folder
+        name = name or folder.name
+
+        if not (isinstance(cls, type) and issubclass(cls, NMRSpectrum)):
```
ScottSoren: would `issubclass(cls, NMRSpectrum)` ever evaluate to True and `isinstance(cls, type)` evaluate to False?

src/ixdat/readers/bruker.py
```diff
@@ -0,0 +1,279 @@
...
+                back to :class:`NMRSpectrum`.
+            procno (int): Which processed-data subfolder to read from
+                (``pdata/<procno>``). Defaults to 1.
+            prefer_processed (bool): If True (default) and a processed
+                spectrum is available, return it. If False, or if no processed
+                data is found, return the magnitude of the raw FID instead.
+            kwargs: Forwarded to ``cls``.
+        """
+        try:
+            import nmrglue as ng
```
ScottSoren, line 136: I'd argue the abbreviation isn't necessary, but I see that's what they do in their documentation, so fine!

src/ixdat/readers/bruker.py
```diff
@@ -0,0 +1,279 @@
...
+                pass
+            # The raw FID (Free Induction Decay) is the time-domain signal: the
+            # exponentially decaying oscillations from precessing nuclear spins,
+            # recorded after an RF pulse. It is complex (two receiver channels
+            # shifted 90 degrees apart). Without Fourier transform + phase
+            # correction + chemical-shift referencing we cannot assign ppm
+            # positions, so we return the magnitude (envelope) against a plain
+            # point index instead.
+            y = np.abs(np.asarray(data))
+            x_ppm = np.arange(y.shape[-1], dtype=float)
```
ScottSoren: Hm... this of course gives `x_ppm = [0, 1, 2, 3, 4, ...]`, which is also what I see when setting `prefer_processed=False` in the demo script (I think this should be demo'd there!) and then `spec.plot()`.
But isn't the "time" of the time domain data available somewhere? Could it be reconstructed from the metadata to be time in (milli? micro? nano?)-seconds?

src/ixdat/readers/bruker.py
```diff
@@ -0,0 +1,279 @@
...
+    except (TypeError, ValueError):
+        pass
+    try:
+        dt = datetime.datetime.fromisoformat(str(date))
+        return dt.timestamp()
+    except ValueError:
+        return None
+
+
+def _to_jsonable(obj):
```
ScottSoren: this seems more general. Perhaps it should go in ixdat/src/ixdat/tools.py?

src/ixdat/techniques/nmr.py
```diff
@@ -0,0 +1,35 @@
...
+    Chemical shift (ppm) expresses a nucleus's resonance frequency relative to a
+    reference compound: delta = (freq_sample - freq_ref) / freq_ref * 1e6. Using
+    parts per million makes peak positions independent of the spectrometer field
+    strength, so the same compound looks the same whether measured at 400 MHz or
+    700 MHz. By NMR convention the ppm axis runs right-to-left (high ppm on the left).
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.plotter = SpectrumPlotter(spectrum=self)
```
ScottSoren: Nice! If you're making a custom technique class, you may as well make a custom plotter to go with it ;)

ScottSoren (changes requested):
Nice work! 
Just a few questions and small changes requested here. The biggest is a request for an NMRPlotter that does the xasix inversion so the user doesn't have to (may as well), and a request to build a physically meaningful xseries based on the acquisition parameters for non-processed NMR (nice to have, not need to have).
I agree that the lazy import works fine in this case :)

src/ixdat/readers/bruker.py
FrederikLizakJohansen: You're right, no reason to hide these. Addressed in 83a52a03c2876a5b6b7f0c7eb225970ad85a3fc0

development_scripts/reader_demonstrators/demo_bruker_reader.py
FrederikLizakJohansen: Yes! Addressed in b6fe8233c14fb7d3543e167681ee9173eea2d425

src/ixdat/readers/bruker.py
FrederikLizakJohansen: I think so too :))

test_data/bruker/MTBLS1_ADG19007u_162_10/pdata/1/proc
FrederikLizakJohansen: ;-)

src/ixdat/readers/bruker.py
FrederikLizakJohansen: I don't see a clean way to return both in the current ixdat framework; so two explicit call would be the easiest. However I do believe the name of the parameter could be more explicit and its behavior as well. In ac9497e69721142758a09a8e3a8e0b79086911d4 and baffa9e8abe7d039efff265533e10873f83f9c76, I renamed "prefer_processed" to "processed" and removed the silent fallback and made it raise an error instead if the processed data isn't there.

development_scripts/reader_demonstrators/demo_bruker_reader.py
FrederikLizakJohansen: You're right! I have added an NMRPlotter in 9de726f9628ea2d0870959e262e20eb4d570d900

src/ixdat/readers/bruker.py
FrederikLizakJohansen: You're right, that combination is actually impossible. I was trying to prevent the TypeError in a clumsy way. Addressed in e8b9f8e95a96f8c3d45c3df601e76a2e784652f2, with an explicit try/except with a fallback to NMRSpectrum if the user parses anything that is not the class or subclass of NMRSpectrum.

src/ixdat/readers/bruker.py
FrederikLizakJohansen: True, but yeah just following the docs :-)

src/ixdat/readers/bruker.py
FrederikLizakJohansen: Good point! I also now realized that a raw FID and a processed NMR spectrum are fundamentally different things right, time-domain vs frequency-domain, so it didn't make sense to plot them the same way. I ended up adding a `FIDSpectrum` class that uses a plain plotter without the x-axis inversion. The reader now returns `NMRSpectrum` for processed data and `FIDSpectrum` for the raw FID. Changed in 3bb4fe88df86f2457a3cd0faef023cd1e50a0f22

src/ixdat/readers/bruker.py
FrederikLizakJohansen: Good point. Moved it to `tools.py`. No other reader uses it yet but it's the right home for it. See 24213259b302defe736a31fb581494c353098368

src/ixdat/techniques/nmr.py
FrederikLizakJohansen: Yes I now agree! Implemented as per the other comments :-)

src/ixdat/readers/bruker.py
ScottSoren: Ah, makes sense to me now! Either way is fine. I can see that in other readers I didn't check that cls is a class, so that resulting type-error would crash the program (which is also fine).

src/ixdat/readers/bruker.py
ScottSoren: Looks good!

src/ixdat/readers/bruker.py
```diff
@@ -0,0 +1,272 @@
...
+            # comes from Nyquist — the ADC samples at twice the spectral width.
+            sw_h = dic["acqus"]["SW_h"]
+            x_ppm = np.arange(y.shape[-1]) / (2.0 * sw_h)
+
+        self.dic = dic
+        metadata = self._build_metadata(dic, processed=processed)
+        tstamp = _extract_tstamp(dic)
+
+        xseries = DataSeries(
+            name="chemical shift" if processed else "time",
```
ScottSoren: Nice! Tiny thing: can you call the xseries of an FIDSpectrum "time / s" so it also appears in the default plot?

ScottSoren (approved):
Looks great! 
A trivial comment, otherwise ready to merge. I can see that there are merge conflicts - also look trivial. Can you merge it into main?

src/ixdat/readers/bruker.py
FrederikLizakJohansen: Yes true, addressed in 3e695d58286addcf8907b4112c4f936c3995137d

PR #203, opened 2026-04-20
Add XRDXYReader for .xy/.xye/.dat X-ray diffraction files

FrederikLizakJohansen:
Adds `XRDXYReader`, a simple reader for the generic two- or three-column powder diffraction text format that a lot of XRD software exports.

## What it does

- Reads two-column `.xy` ($2\theta$ or Q, intensity) and three-column `.xye` (+ error) files
- Also handles comma-separated files
- Scans comment lines for column labels to figure out whether it's $2\theta$ or Q-space, and picks up the units (`degree`, `1/angstrom`, `1/nm`)
- Falls back to `two theta / degree` when there's no header
- Stores the error column in `metadata["intensity_error"]` for `.xye` files
- Registered as `"xy"` in `SPECTRUM_READER_CLASSES`,  so `Spectrum.read(path, reader="xrdxy")`

## What it doesn't do

No auto-detection from file suffix: `.xy` is too generic for that. Users have to pass `reader="xrdxy"` explicitly, which is consistent with how other spectrum readers work.

## Files

- `src/ixdat/readers/xrd_xy.py`: The reader
- `src/ixdat/readers/__init__.py`: Registration
- `development_scripts/reader_demonstrators/demo_xrd_xy_reader.py`: Demo using ZSM-5 synchrotron data (`.xye`, no header) and rutile TiO2 I(Q) from DebyeCalculator (`.dat`, `Q,I(Q)` header, comma-separated)


development_scripts/reader_demonstrators/demo_xrd_xy_reader.py
```diff
@@ -0,0 +1,53 @@
...
+"""
+
+import urllib.request
+from pathlib import Path
+
+from ixdat import Spectrum
+
+import matplotlib.pyplot as plt
+
+DATA_DIR = Path(__file__).parent / "xrd_xy_data"
```
ScottSoren: Really nice, using the data that you already have on an online reposotory :) . Only problem is that I want to avoid having data end up in ixdat/development_scripts which risks appearing/dissapearing by mistake in commits. Could you do one of the following?
- make a `Spectrum.read_url` which does exactly the same as [Measurement.read_url](https://github.com/ixdat/ixdat/blob/26d5d2ebab53a0e27b822e9d480a78f05c88bb94/src/ixdat/measurement_base.py#L295)
- move the data files to ixdat/test_data/xrd_xy_data as a permanent part of ixdat
- make DATA_DIR a temporary directory
- add something to .gitignore so that the data files won't get committed

I think of the above that `Spectrum.read_url` is the best, since it would be most generally useful :)

src/ixdat/readers/xrd_xy.py
```diff
@@ -0,0 +1,159 @@
...
+        """Read an .xy or .xye file and return a Spectrum.
+
+        Args:
+            path_to_file (str or Path): Path to the file.
+            cls (Spectrum subclass): Class to instantiate. Defaults to Spectrum.
+            kwargs: Passed to cls.from_field (e.g. name, sample_name).
+
+        Returns:
+            Spectrum with x (2-theta or Q) and y (intensity) series. For .xye
+            files the error column is stored in ``metadata["intensity_error"]``.
```
ScottSoren: Hm... you've stored data as if it was metadata. I don't think that's a good idea.
To do it right, I think reading a .xye should actually result in a  [MultiSpectrum](https://github.com/ixdat/ixdat/blob/26d5d2ebab53a0e27b822e9d480a78f05c88bb94/src/ixdat/spectra.py#L332) object.
A MultiSpectrum is now just used to get `Spectrum` objects by indexing. But you could make a new class in a techniques/xrd.py module which inherits from MultiSpectrum and plots y vs x by default - or, better, also make a new plotter which plots y+e and y-e vs x in a nice-looking way :)

src/ixdat/readers/xrd_xy.py
```diff
@@ -0,0 +1,159 @@
...
+
+# Exact lowercased column-header tokens that unambiguously indicate Q-space.
+# Written by software such as GSAS-II, FullProf, and DebyeCalculator.
+_Q_TOKENS = {"q", "q(a^-1)", "q(nm^-1)", "q/a^-1", "q/nm^-1"}
+
+# Tokens that indicate a 2-theta x axis (angle in degrees).
+_TWOTHETA_TOKENS = {"2theta", "2-theta", "2th", "angle"}
+
+
+def _parse_header_and_data(path_to_file):
```
ScottSoren, line 17: Why don't you use `Pandas.read_csv`? Pandas is already a dependency.

ScottSoren (changes requested):

development_scripts/reader_demonstrators/demo_xrd_xy_reader.py
FrederikLizakJohansen: Good call, I've added `Spectrum.read_url` and updated the demo to use it directly in d72e59606e868a5821f5d64597dda767c3854312

src/ixdat/readers/xrd_xy.py
FrederikLizakJohansen: Good catch, I did not think about this. I've added `XRDSpectrum` (inherits from `MultiSpectrum`). The reader now always returns `XRDSpectrum` for both `.xy` and `.xye` files. For `.xye` the second field is the per-point error, and a new `XRDSpectrumPlotter` draws a shaded y+e, y-e band. The ZSM-5 dataset in the demo is already a `.xye `file, so you can see the band in that (although the error is tiny). Change in 8199d34bf5bebbaca1de8aff81d508031ebf0b8a.

Perhaps the xrdml files reader should also eventually use this same Spectrum class? This ofc would be a separate PR. I could open an issue if you agree?

src/ixdat/readers/xrd_xy.py
FrederikLizakJohansen: Fair point. Switched to `pd.read_csv` with `sep=r'\s+|,'` which handles both space and comma delimiters automatically, removing the manual `data_lines` accumulation, delimiter detection, and `np.loadtxt.` We still need the pre-scan to collect header/label lines (since pandas comment only takes one character), but the data loading itself is now simpler. Changes in 1c7f408bd3569de3cdfe2890789e1f0b98c85551

src/ixdat/plotters/xrd_plotter.py
```diff
@@ -0,0 +1,33 @@
+"""Plotter for XRD spectra."""
+
+from .base_mpl_plotter import MPLPlotter
+
+
+class XRDSpectrumPlotter(MPLPlotter):
```
ScottSoren, line 6: Hm... this doesn't actually have anything to do with XRD. 
It could be called a `SpectrumCIPlotter` with CI standing for "Confidence Interval". But it's fine so long as the only data we have of the x-y-e type is XRD. 
I ran the demo and zoomed in - it works very nicely :)

src/ixdat/readers/xrd_xy.py
ScottSoren: Hm... I'm a bit confused though since the demo [xy file](https://raw.githubusercontent.com/FrederikLizakJohansen/DebyeCalculator/main/debyecalculator/unittests_files/icsd_001504_cc_r6_lc_2.85_6_tetragonal_Iq.dat) has a header "Q,I(Q)" that pandas should have no problem with; and the demo [xye file](https://raw.githubusercontent.com/stefsmeets/lines/main/testing/zsm-5.xye) looks like it has no headers at all.
Do you have a demo file for which the necessity of this code is clear?
No big problem, though; I can see how it also works fine for the demo files and can imagine the additional types of headers it accomodates.

src/ixdat/readers/xrd_xy.py
```diff
@@ -0,0 +1,167 @@
...
+
+    1/angstrom is the crystallography convention; 1/nm is common in
+    small-angle scattering and simulation tools.
+    """
+    if "nm" in token:
+        return "1/nm"
+    return "1/angstrom"
+
+
+def _is_numeric(s):
```
ScottSoren: This could go in tools.py

src/ixdat/readers/xrd_xy.py
```diff
@@ -0,0 +1,167 @@
...
+        xseries = DataSeries(name=x_name, unit_name=x_unit, data=data[:, 0])
+        intensity_field = Field(
+            name=y_name, unit_name=y_unit, data=data[:, 1], axes_series=[xseries]
+        )
+
+        from ..techniques.xrd import XRDSpectrum
+
+        try:
+            if not issubclass(cls, XRDSpectrum):
+                cls = XRDSpectrum
```
ScottSoren, line 146: hm... I wonder if it will ever cause a problem for a user expecting a `Spectrum` to get a `MultiSpectrum`? I think basically never, because all you'd use in a `Spectrum` is `x`, `y` and `plot`, and those function as normal. The one exception is if you have a bunch of diffractograms taken right after each other, and try to add them. If they were `Spectrum` objects, then `xrd_1 + xrd_2 + xrd_3...` would give a `SpectrumSeries` object with the powerful plotters that comes with. The `+` operator is not defined in a `MultiSpectrum`.

Let's wait until someone actually wants to use it like that. This is good for now :)

src/ixdat/techniques/xrd.py
```diff
@@ -0,0 +1,36 @@
+"""Module for representation and analysis of XRD spectra."""
+
+from ..spectra import MultiSpectrum
+from ..plotters.xrd_plotter import XRDSpectrumPlotter
+
+
+class XRDSpectrum(MultiSpectrum):
```
ScottSoren, line 7: This doesn't have anything to do with XRD - maybe it's a "ConfidenceIntervalSpectrum" or something like that, from which XRDSpectrum could inherit.
but it's fine for now.

ScottSoren (approved):
Looks great! The implementation inspired some more questions and thoughts for possible future development, in the comments here. But looks good to merge :D 
Nice work!

src/ixdat/plotters/xrd_plotter.py
FrederikLizakJohansen: Yeah fair point, the error band logic is totally generic. XRD is the only x-y-e type we have right now so the name works for the moment. If that changes we can refactor then, I opened an issue to track it: #204

src/ixdat/readers/xrd_xy.py
FrederikLizakJohansen: Agreed, no point solving that until someone actually runs into it. Opened an issue so we do not forget: #205

src/ixdat/techniques/xrd.py
FrederikLizakJohansen: Agreed; see other comment, and #204

src/ixdat/readers/xrd_xy.py
FrederikLizakJohansen: Done, moved it in dd93b97a3d3adc9254850054cb2549882ce4d360

src/ixdat/readers/xrd_xy.py
FrederikLizakJohansen: You are totally right to question this; I should not be missing these cases. I am making some tests to be  sure that the reader parses the data correctly.

FrederikLizakJohansen: I have added a bunch of tests in the commits below; and actually found a severe case where axis labels preceded by comments would cause a parsing-error. That is fixed in 8f0daf4a6acb4472754661acc1c7efb5fab53c36. No need for further review on the tests. Thank you for making me aware of this!

Issue #206, opened 2026-06-19
AsimovReader needs to support retrieving Spectrum-class data from ASIMOV.enci.dk


FrederikLizakJohansen:
Currently AsimovReader is in the list of READER_CLASSES, but not SPECTRUM_READER_CLASSES.
This, and probably some additional supporting factors, means that a user is currently unable to access data from ASIMOV via an ID of a Spectrum object.

PR #207, opened 2026-06-24
Add ASIMOV support for Spectrum payloads

FrederikLizakJohansen:
# Support Spectrum and SpectrumSeries payloads in ASIMOV reader

This PR addresses ixdat issue #206: ASIMOV should be able to return both Measurement-class and Spectrum-class data through the normal ixdat read API.

The important user-facing goal is that these calls should all work when the ASIMOV payload contains the matching ixdat object, including registered subclasses of either class:

```python
Measurement.read(asimov_id, reader="asimov")
Spectrum.read(asimov_id, reader="asimov")
```

There are three parts to making that work cleanly.

First, ixdat needed the same reconstruction behavior for `Spectrum` that it already had for `Measurement`. `Measurement.from_dict()` already looks at the payload technique and returns the registered technique-specific class. Spectrum did not do that before, so a technique-specific Spectrum payload could be reconstructed as plain `Spectrum` instead of the registered subclass. This PR adds `Spectrum.from_dict()` so spectrum payloads get the correct subclass, plotter, and behavior.

Second, ASIMOV has migrated from dataset/dataset-version payload storage to project-file and bundle payload storage. The reader therefore had to stop using the legacy compatibility routes and start using the native ixdat-payload endpoints.

Third, ASIMOVReader should not need a new special case every time we add a new technique, spectrum subtype, or reader. The reader now builds ixdat's core container objects generically from `object_type`, delegates technique dispatch to ixdat's own `from_dict()` methods, and delegates series class selection to ixdat's `DataSeries.from_dict()`.

### `Spectrum.from_dict()` now mirrors `Measurement.from_dict()`

`src/ixdat/spectra.py`

Minimal excerpt:

```python
@classmethod
def from_dict(cls, obj_as_dict):
    from .techniques import TECHNIQUE_CLASSES

    technique_class = TECHNIQUE_CLASSES.get(obj_as_dict.get("technique"))
    if technique_class and issubclass(technique_class, cls):
        return technique_class(**obj_as_dict)
    return cls(**obj_as_dict)
```

Why this matters:

ASIMOV can now send a spectrum payload with `"technique": "NMR"` and ixdat can rebuild it as `NMRSpectrum`, not just `Spectrum`. That is important because the subclass carries technique-specific plotting behavior.

### ASIMOV is registered as a Spectrum reader

`src/ixdat/readers/__init__.py`

Minimal excerpt:

```python
SPECTRUM_READER_CLASSES = {
    ...
    "bruker": BrukerNMRReader,
    "asimov": AsimovReader,
}
```

Why this matters:

`Spectrum.read(..., reader="asimov")` looks in `SPECTRUM_READER_CLASSES`. Without this registration, the Measurement path could work while the Spectrum path could not even find the ASIMOV reader.

### ASIMOVReader uses the native project-file and bundle payload endpoints

`src/ixdat/readers/asimov.py`

Minimal excerpt:

```python
payload_envelope = self._get_ixdat_payload_envelope(
    asimov_id, headers=headers
)
```

```python
def _get_ixdat_payload_envelope(self, asimov_id, headers):
    try:
        return self._get_json(
            f"project-files/{asimov_id}/ixdat-payload", headers=headers
        )
    except RuntimeError as exc:
        if not self._is_http_status(exc, 404):
            raise

    try:
        return self._get_json(
            f"project-file-bundles/{asimov_id}/ixdat-payload", headers=headers
        )
    ...
```

Why this matters:

The old ASIMOV compatibility routes returned a legacy dataset-version envelope with synthetic `dataset_id` and `version` fields. The new native endpoints return the clean ixdat payload envelope for a project file or bundle. ixdat should read that native envelope directly.

### ASIMOV provenance now uses project-file and bundle fields

`src/ixdat/readers/asimov.py`

Minimal excerpt:

```python
meta = dict(payload.get("metadata") or {})
meta["asimov"] = {
    "project_file_id": payload_envelope.get("project_file_id"),
    "bundle_id": payload_envelope.get("bundle_id"),
    "filename": payload_envelope.get("filename"),
    "name": payload_envelope.get("name"),
    "parser_name": payload_envelope.get("parser_name"),
    "created_at": payload_envelope.get("created_at"),
    "ixdat_version": payload_envelope.get("ixdat_version"),
}
```

Why this matters:

Users still need provenance, but it should describe the current ASIMOV storage model. The metadata should not preserve legacy dataset-version concepts that are only present in compatibility shims.

### The reader catches wrong read entrypoints early

`src/ixdat/readers/asimov.py`

Minimal excerpt:

```python
object_type = d.get("object_type", "measurement")
self._check_requested_class_matches_object_type(
    requested_class=cls, object_type=object_type, asimov_id=asimov_id
)
```

```python
if object_type == "measurement":
    compatible = issubclass(requested_class, Measurement)
elif object_type == "spectrum":
    compatible = issubclass(requested_class, Spectrum) and not issubclass(
        requested_class, SpectrumSeries
    )
elif object_type == "spectrum_series":
    compatible = issubclass(requested_class, Spectrum)
```

Why this matters:

If the user calls `Spectrum.read(..., reader="asimov")` on a Measurement payload, the useful answer is "this is a Measurement payload; use `Measurement.read()`". Without this check, the user would get a confusing constructor error such as `unexpected keyword argument 'series_list'`.

### Container objects are built from `object_type`

`src/ixdat/readers/asimov.py`

Minimal excerpt:

```python
OBJECT_TYPE_CLASSES = {
    "measurement": Measurement,
    "spectrum": Spectrum,
    "spectrum_series": SpectrumSeries,
}
```

```python
def _build_object(self, dct, cls=None, **kwargs):
    object_type = dct.get("object_type", "measurement")
    if cls is None:
        cls = OBJECT_TYPE_CLASSES.get(object_type)
        ...
    if cls is Spectrum and object_type == "spectrum_series":
        cls = SpectrumSeries
    return cls.from_dict(self._build_kwargs(dct, object_class=cls, **kwargs))
```

Why this matters:

ASIMOVReader should not know about every technique. It only needs to know the core ixdat container family: Measurement, Spectrum, or SpectrumSeries. The technique-specific class is then selected by `from_dict()`.

### Nested ixdat objects are hydrated generically

`src/ixdat/readers/asimov.py`

Minimal excerpt:

```python
if key not in obj and isinstance(value, dict) and "object_type" in value:
    obj[key] = self._build_object(value, reader=self)
elif (
    key not in obj
    and isinstance(value, list)
    and any(isinstance(v, dict) and "object_type" in v for v in value)
):
    obj[key] = [
        self._build_object(v, reader=self)
        if isinstance(v, dict) and "object_type" in v
        else v
        for v in value
    ]
```

Payload shape:

```python
{
    "object_type": "measurement",
    "technique": "EC-Optical",
    "spectrum_series": {"object_type": "spectrum_series", ...},
    "reference_spectrum": {"object_type": "spectrum", ...},
}
```

Why this matters:

This avoids special-casing names like `reference_spectrum`. The rule is simple:
if a nested value is an ixdat object, ASIMOV marks it with `object_type`, and ixdat rebuilds it recursively.

### Nested ixdat objects must declare `object_type`

`src/ixdat/readers/asimov.py`

Minimal excerpt:

```python
elif (
    key not in obj
    and isinstance(value, dict)
    and "object_type" not in value
    and ("series_list" in value or "field" in value)
):
    raise ValueError(
        f"Asimov payload field {key!r} is missing 'object_type'. "
        "Nested ixdat objects must declare object_type so ixdat can "
        "reconstruct them without key-specific reader logic."
    )
```

Bad payload:

```python
"reference_spectrum": {
    "name": "reference",
    "technique": "spectrum",
    "field": {...},
}
```

Good payload:

```python
"reference_spectrum": {
    "object_type": "spectrum",
    "name": "reference",
    "technique": "spectrum",
    "field": {...},
}
```

Why this matters:

The reader should not guess object type from a key name or from array shape. That would make future payloads fragile. The payload itself should say what each nested ixdat object is.

### Technique-specific linker fields are preserved when ixdat declares them

`src/ixdat/readers/asimov.py`

Minimal excerpt:

```python
for key in self._get_serializable_attrs(dct, object_class):
    if key not in obj and key in dct:
        obj[key] = dct[key]
```

```python
serializable_attrs = set(object_class.get_all_column_attrs())
...
if technique_class and issubclass(technique_class, object_class):
    serializable_attrs.update(technique_class.get_all_column_attrs())
```

Why this matters:

Some technique classes can be built either from embedded child objects or saved object IDs. For EC-Optical data, `ref_id` and `spectrum_id` are meaningful ixdat constructor fields. The reader passes those through only when ixdat declares them, instead of passing arbitrary ASIMOV metadata into constructors.

### DataSeries construction uses ixdat's series registry

`src/ixdat/readers/asimov.py`

Minimal excerpt:

```python
kind = dct.get("series_type", "series")
series_cls = SERIES_CLASSES.get(kind)
...
return DataSeries.from_dict(series_dict)
```

```python
if "tseries_key" in dct and key_map:
    series_dict["tseries"] = key_map[dct["tseries_key"]]
if "axes_keys" in dct and key_map:
    series_dict["axes_series"] = [key_map[k] for k in dct["axes_keys"]]
```

Why this matters:

ASIMOVReader still resolves ASIMOV-specific references like `tseries_key` and `axes_keys`, but ixdat itself chooses the DataSeries subclass. This means a new registered ixdat series type does not require a new ASIMOVReader special case.

### ASIMOV-only series metadata is filtered out

`src/ixdat/readers/asimov.py`

Minimal excerpt:

```python
serializable_attrs = set(series_cls.get_all_column_attrs())
series_dict = {
    key: value for key, value in dct.items() if key in serializable_attrs
}
```

Payload shape:

```python
{
    "series_type": "tseries",
    "class": "TimeSeries",
    "name": "time/s",
    "unit_name": "s",
    "data": [0.0, 1.0],
    "tstamp": 1.0,
}
```

Why this matters:

Real ASIMOV series payloads can contain helper metadata such as `"class"`. Python constructors reject unknown keyword arguments, so the reader only passes fields that ixdat declares as constructor data.

### `development_scripts/reader_demonstrators/inspect_asimov_payloads.py`

Added a small inspection script for looking at ASIMOV payload envelopes and payload structure without dumping full numerical arrays.

Minimal excerpt:

```python
def _get_payload(reader, asimov_id, force_login=False):
    headers = reader._build_auth_headers(force_login=force_login)
    envelope = reader._get_ixdat_payload_envelope(asimov_id, headers=headers)
    payload = envelope.get("payload_json")
    if not payload and envelope.get("payload_uri"):
        payload = reader._load_payload_uri(envelope["payload_uri"], headers=headers)
    return envelope, payload
```

Why this matters:

While developing this PR, ASIMOV payloads differed across techniques and changed as the ASIMOV-side migration progressed. The helper makes it easy to inspect payload shape without printing huge arrays.

## Tests

The tests were expanded to cover the behavior described above: native ASIMOV endpoints, Measurement/Spectrum mismatch errors, Spectrum and SpectrumSeries payloads, NMR subclass dispatch, nested ixdat objects, missing nested `object_type`, linker IDs such as `ref_id`, and ASIMOV-only series metadata.

### DIFF
--- a/development_scripts/reader_demonstrators/inspect_asimov_payloads.py
+++ b/development_scripts/reader_demonstrators/inspect_asimov_payloads.py
@@ -1,0 +1,124 @@
+"""Inspect current Asimov payload structure without printing full data arrays.
+
+Usage:
+    python development_scripts/reader_demonstrators/inspect_asimov_payloads.py ID [ID ...]
+
+Set ASIMOV_ACCESS_TOKEN for non-interactive use, or let AsimovReader use the
+normal Keycloak login flow.
+"""
+
+import argparse
+from collections import Counter
+
+from ixdat.readers.asimov import AsimovReader
+
+
+def _shape(value):
+    """Return the nested list shape if value looks like array data."""
+    shape = []
+    cursor = value
+    while isinstance(cursor, list):
+        shape.append(len(cursor))
+        cursor = cursor[0] if cursor else None
+    return tuple(shape)
+
+
+def _short(value):
+    """Return a compact scalar/list summary suitable for terminal inspection."""
+    if isinstance(value, list):
+        return f"list shape={_shape(value)}"
+    if isinstance(value, dict):
+        return f"dict keys={sorted(value)}"
+    return repr(value)
+
+
+def _print_series(entry, indent):
+    prefix = " " * indent
+    print(
+        f"{prefix}- {entry.get('series_type', 'series')}: "
+        f"{entry.get('name')!r} [{entry.get('unit_name')!r}]"
+    )
+    if "key" in entry:
+        print(f"{prefix}  key={entry['key']!r}")
+    if "data" in entry:
+        print(f"{prefix}  data={_short(entry['data'])}")
+    for ref_key in ("tseries_key", "axes_keys"):
+        if ref_key in entry:
+            print(f"{prefix}  {ref_key}={entry[ref_key]!r}")
+    if entry.get("axes_series"):
+        print(f"{prefix}  axes_series:")
+        for axis in entry["axes_series"]:
+            _print_series(axis, indent + 4)
+
+
+def _print_object(dct, indent=0, label="payload"):
+    prefix = " " * indent
+    object_type = dct.get("object_type", "<missing>")
+    technique = dct.get("technique", "<missing>")
+    print(f"{prefix}{label}: object_type={object_type!r}, technique={technique!r}")
+    for key in ("name", "sample_name", "tstamp", "duration", "durations", "continuous"):
+        if key in dct:
+            print(f"{prefix}  {key}={_short(dct[key])}")
+    if isinstance(dct.get("metadata"), dict):
+        print(f"{prefix}  metadata keys={sorted(dct['metadata'])}")
+    if dct.get("series_list"):
+        counts = Counter(s.get("series_type", "series") for s in dct["series_list"])
+        print(f"{prefix}  series_list len={len(dct['series_list'])} types={dict(counts)}")
+        for series in dct["series_list"][:6]:
+            _print_series(series, indent + 4)
+        if len(dct["series_list"]) > 6:
+            print(f"{prefix}    ... {len(dct['series_list']) - 6} more series")
+    if isinstance(dct.get("field"), dict):
+        print(f"{prefix}  field:")
+        _print_series(dct["field"], indent + 4)
+    nested = [
+        (key, value)
+        for key, value in dct.items()
+        if isinstance(value, dict) and "object_type" in value
+    ]
+    for key, value in nested:
+        _print_object(value, indent + 2, label=key)
+
+
+def _get_payload(reader, asimov_id, force_login=False):
+    headers = reader._build_auth_headers(force_login=force_login)
+    envelope = reader._get_ixdat_payload_envelope(asimov_id, headers=headers)
+    payload = envelope.get("payload_json")
+    if not payload and envelope.get("payload_uri"):
+        payload = reader._load_payload_uri(envelope["payload_uri"], headers=headers)
+    return envelope, payload
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("asimov_ids", nargs="+")
+    parser.add_argument("--force-login", action="store_true")
+    args = parser.parse_args()
+
+    reader = AsimovReader()
+    for asimov_id in args.asimov_ids:
+        envelope, payload = _get_payload(
+            reader,
+            asimov_id,
+            force_login=args.force_login,
+        )
+        print("=" * 88)
+        print(
+            f"project_file_id={envelope.get('project_file_id')} "
+            f"bundle_id={envelope.get('bundle_id')} "
+            f"filename={envelope.get('filename')!r} "
+            f"name={envelope.get('name')!r}"
+        )
+        print(
+            f"created_at={envelope.get('created_at')} "
+            f"parser={envelope.get('parser_name')!r} "
+            f"ixdat={envelope.get('ixdat_version')!r}"
+        )
+        if payload:
+            _print_object(payload)
+        else:
+            print("payload: <missing>")
+
+
+if __name__ == "__main__":
+    main()
--- a/development_scripts/reader_demonstrators/inspect_asimov_payloads.py
+++ b/development_scripts/reader_demonstrators/inspect_asimov_payloads.py
@@ -1,7 +1,7 @@
 """Inspect current Asimov payload structure without printing full data arrays.
 
 Usage:
-    python development_scripts/reader_demonstrators/inspect_asimov_payloads.py ID [ID ...]
+    python inspect_asimov_payloads.py ID [ID ...]
 
 Set ASIMOV_ACCESS_TOKEN for non-interactive use, or let AsimovReader use the
 normal Keycloak login flow.
@@ -63,7 +63,9 @@ def _print_object(dct, indent=0, label="payload"):
         print(f"{prefix}  metadata keys={sorted(dct['metadata'])}")
     if dct.get("series_list"):
         counts = Counter(s.get("series_type", "series") for s in dct["series_list"])
-        print(f"{prefix}  series_list len={len(dct['series_list'])} types={dict(counts)}")
+        print(
+            f"{prefix}  series_list len={len(dct['series_list'])} types={dict(counts)}"
+        )
         for series in dct["series_list"][:6]:
             _print_series(series, indent + 4)
         if len(dct["series_list"]) > 6:
--- a/src/ixdat/readers/asimov.py
+++ b/src/ixdat/readers/asimov.py
@@ -1,4 +1,4 @@
-"""Reader for ixdat measurements stored in Asimov dataset versions."""
+"""Reader for ixdat objects stored as parsed Asimov project-file payloads."""
 
 import os
 from dataclasses import dataclass, fields
@@ -8,7 +8,7 @@ import numpy as np
 import requests
 
 from ..auth import KeycloakDeviceTokenProvider
-from ..data_series import DataSeries, TimeSeries, ValueSeries, Field
+from ..data_series import DataSeries, SERIES_CLASSES
 from ..measurement_base import Measurement
 from ..spectra import Spectrum, SpectrumSeries
 from ..tools import request_with_retries
@@ -111,9 +111,9 @@ ASIMOV_CONFIG = AsimovConfig.from_env()
 
 
 class AsimovReader:
-    """Read a remote dataset version from Asimov into an ixdat object.
+    """Read a remote project file or bundle parse result into an ixdat object.
 
-    Typical use only needs the dataset id: ``Measurement.read(<id>,
+    Typical use only needs the project file or bundle id: ``Measurement.read(<id>,
     reader="asimov")``. The first read opens a browser tab where you log
     in once; later reads reuse a cached token.
 
@@ -159,8 +159,7 @@ class AsimovReader:
                 read_timeout=self.config.read_timeout,
             )
 
-        self.dataset = None
-        self.dataset_version = None
+        self.payload_envelope = None
         self._session = requests.Session()
         self._session.trust_env = False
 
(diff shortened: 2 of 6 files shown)


development_scripts/reader_demonstrators/inspect_asimov_payloads.py
```diff
@@ -0,0 +1,126 @@
+"""Inspect current Asimov payload structure without printing full data arrays.
+
+Usage:
+    python inspect_asimov_payloads.py ID [ID ...]
+
+Set ASIMOV_ACCESS_TOKEN for non-interactive use, or let AsimovReader use the
+normal Keycloak login flow.
+"""
+
+import argparse
```
ScottSoren, line 10: I guess it may be part of coming discussion connected to #212 , but it'd perhaps nice to have in the reader_demnstrators folder a version of this that runs without args, using pre-selected demo data... or adding spectrum stuff to demo_asimov_readery, and put this script a folder up.

development_scripts/reader_demonstrators/inspect_asimov_payloads.py
```diff
@@ -0,0 +1,126 @@
...
+from collections import Counter
+
+from ixdat.readers.asimov import AsimovReader
+
+
+def _shape(value):
+    """Return the nested list shape if value looks like array data."""
```
ScottSoren, line 17: interesting, but why would it not _be_ array data (with a `shape` attribute) if it looks like array data?

development_scripts/reader_demonstrators/inspect_asimov_payloads.py
```diff
@@ -0,0 +1,126 @@
...
+def _short(value):
+    """Return a compact scalar/list summary suitable for terminal inspection."""
+    if isinstance(value, list):
+        return f"list shape={_shape(value)}"
+    if isinstance(value, dict):
+        return f"dict keys={sorted(value)}"
+    return repr(value)
+
+
+def _print_series(entry, indent):
```
ScottSoren, line 35: Nice! This representation actually looks like it should be copied to `DataSeries.__str__`.
As is now, `TimeSeries.__str__` and `ValueSeries.__str__` are implemented show the range of the data they contained, which is useful for `Measurement.__str__`. But that could be refactored to something like `_description`.

development_scripts/reader_demonstrators/inspect_asimov_payloads.py
```diff
@@ -0,0 +1,126 @@
...
+    for ref_key in ("tseries_key", "axes_keys"):
+        if ref_key in entry:
+            print(f"{prefix}  {ref_key}={entry[ref_key]!r}")
+    if entry.get("axes_series"):
+        print(f"{prefix}  axes_series:")
+        for axis in entry["axes_series"]:
+            _print_series(axis, indent + 4)
+
+
+def _print_object(dct, indent=0, label="payload"):
```
ScottSoren, line 54: Now I'm getting a bit confused about the point of this module. Wouldn't you test exactly the same by converting `dct` to an ixdat object (it looks like `Measurement` is what's in mind here) and printing the ixdat object? As per above, `Measurement.__str__` or `Measurement._description` could be made to return the string printed here, since I agree it's an informative representation!

ScottSoren: I guess I think of an ixdat object as being such a thin and natural wrapper around an asimov payload that looking at the payload itself is more of a software test than a demo. Or, if it is a demo, then it's one to include in the asimov repository/documentation instead!
... But since I'm not looking at asimov code yet, this may also be a fine place to show it to me, in which case: "received and approved!" The payload structure looks nice :)

development_scripts/reader_demonstrators/inspect_asimov_payloads.py
ScottSoren: This is for asimov rather than ixdat:
It would be nice if the asimov_id of a measurement was easy to find on asimov! That would enable me to try out this script. I know I've asked for it once before and you put it up, but it seems to have disappeared in later asimov updates. At least I can't find it here, for example: 

Image

ScottSoren: Especially annoying as the dataset id's in demo_asmov_reader.py don't work any more. 
When running 
`python inspect_asimov_payloads.py 53ca3c5f-cd75-406c-8d99-a51c397988d7`
I get, for example:
`ValueError: No Asimov project file or project file bundle with id=53ca3c5f-cd75-406c-8d99-a51c397988d7.`
... and when running demo_asimov_reader.py: 
`RuntimeError: Failed to fetch https://asimov.enci.dk/api/datasets/53ca3c5f-cd75-406c-8d99-a51c397988d7 (HTTP 404 Not Found): Dataset not found`

src/ixdat/readers/asimov.py
```diff
@@ -159,82 +159,138 @@ def __init__(self, *, token=None, token_provider=None, config=None):
...
-        version_id=None,
         force_login=False,
         **kwargs,
     ):
-        """Read a dataset version from Asimov as an ixdat object.
+        """Read an ixdat payload from Asimov.
 
         Args:
-            id (str): Asimov dataset id (UUID string).
+            id (str): Asimov project-file or project-file-bundle id (UUID string).
```
ScottSoren, line 176: What's the difference between a "dataset id", a "project-file" and a "project-file-bundle"?

src/ixdat/readers/asimov.py
```diff
@@ -159,82 +159,138 @@ def __init__(self, *, token=None, token_provider=None, config=None):
...
+            requested_class=cls, object_type=object_type, asimov_id=asimov_id
+        )
+
+        return self._build_object(d, cls=cls, reader=self, **kwargs)
+
+    @staticmethod
+    def _check_requested_class_matches_object_type(
+        requested_class, object_type, asimov_id
+    ):
+        """Raise a clear error for the wrong read entrypoint."""
```
ScottSoren, line 232: Would't it be cleaner to have a high-level helper function `get_object_type(cls)` and then just compare strings here?

src/ixdat/readers/asimov.py
```diff
@@ -159,82 +159,138 @@ def __init__(self, *, token=None, token_provider=None, config=None):
...
+            raise
 
-        return cls.from_dict(self._build_kwargs(d, reader=self, **kwargs))
+    @staticmethod
+    def _is_http_status(exc, status_code):
+        """Return True if a request RuntimeError mentions an HTTP status code."""
+        return f"HTTP {status_code}" in str(exc)
 
-    def _build_kwargs(self, dct, **kwargs):
+    def _build_kwargs(self, dct, object_class=None, **kwargs):
```
ScottSoren, line 293: Even though this is a hidden method, could you mention + `object_class` in the docstring?

src/ixdat/readers/asimov.py
```diff
@@ -284,8 +340,83 @@ def _build_kwargs(self, dct, **kwargs):
                 obj["durations"] = dct.get("durations")
                 obj["continuous"] = dct.get("continuous", False)
 
+        for key, value in dct.items():
+            # A nested dict with object_type is a complete ixdat object payload,
+            # such as a reference_spectrum or spectrum_series inside a Measurement.
+            # Build it recursively without caring what the attribute is called.
```
ScottSoren, line 346: A philosophical question I encounter while reviewing this code: Is asimov a database or a file type? This being a **reader** means that ixdat can treat it as a file type. So what you're doing here is the right thing to do: load everything and build the ixdat object and don't worry too much about linking back to the original data source.

But ixdat should eventually also have an asimov **backend** which treats asimov as a database. In that case, loads can in principle be lazy. As in, a `SpectroMeasurement` could be initiated with just a reference to the `SpectrumSeries` and its various `DataSeries` and where to find them in asimov - as is the case with the sqlite backend. Then those linked objects can get built later as needed.

src/ixdat/spectra.py
```diff
@@ -97,6 +97,21 @@ def __init__(
         self.exporter = SpectrumExporter(spectrum=self)
         self.export = self.exporter.export
 
+    @classmethod
+    def from_dict(cls, obj_as_dict):
```
ScottSoren, line 101: thanks for correcting this omission!

tests/unit/test_asimov_reader.py
```diff
@@ -141,6 +186,364 @@ def test_parse_spectrum_payload():
...
+                    "unit_name": "eV",
+                    "data": x.tolist(),
+                }
+            ],
+        },
+    }
+    session = FakeQueuedSession(
+        [
+            make_json_response(
+                make_payload_envelope(payload, filename="remote-spectrum.dat")
```
ScottSoren, line 215: nice.
A worry though is that this does not test whether asimov actually makes the payloads as ixdat expects. If future asimov changes how it structures the payloads, then not only will the ixdat reader have to change, but also this test. 

Not sure what the "correct" software developer attitude to this is?

ScottSoren (changes requested):
I agree with all the motivation and reasoning in your comment.
The code for the most part looks great! The usual questions and suggestions. 
However, I can't approve it yet because I'm missing information to test it. See my comments to inspect_asimov_payloads.py. In short, the reader only seems useful, and the demos only runable, if the user knows where to find the asimov_id of the measurement they want (which I at present do not).

development_scripts/reader_demonstrators/inspect_asimov_payloads.py
FrederikLizakJohansen: You are completely right that this does not belong as a demo; I will keep the inspections to be asimov-side, which makes the most sense anyway as tests here in ixdat will eventually drift and need updating. I will make a issue requesting the ".\__str\__" dunder method to be updated to something more akin to this printer.

development_scripts/reader_demonstrators/inspect_asimov_payloads.py
FrederikLizakJohansen: See https://github.com/ixdat/ixdat/pull/207#discussion_r4061653369

development_scripts/reader_demonstrators/inspect_asimov_payloads.py
FrederikLizakJohansen: removed in b0a7f25ebc2868aad46eebcdf3e08d7f1ebb1012; issue created here #216

development_scripts/reader_demonstrators/inspect_asimov_payloads.py
FrederikLizakJohansen: Yeah sorry about that; things have drifted a lot since this PR was opened. I've updated `demo_asimov_reader.py` (b0a7f25ebc2868aad46eebcdf3e08d7f1ebb1012) to use two permanent example IDs, so you can run it without finding an ID or passing any arguments:

```bash
python development_scripts/reader_demonstrators/demo_asimov_reader.py
```

It reads the NMR example with `Spectrum.read("300f4dcf-1585-51a6-81bc-f867910efe1a", reader="asimov")` and the Biologic CV with `Measurement.read("fbbf4edb-f288-5e85-bed0-e3bc89edc2e5", reader="asimov")`, then plots both.

I've also updated ASIMOV to show the relevant ID clearly in its interface, so you can find the ID for your own data and use it with `Measurement.read()` or `Spectrum.read()`.

development_scripts/reader_demonstrators/inspect_asimov_payloads.py
FrederikLizakJohansen: Good question. The inspector looked at the numbers as they arrived from asimov. At that point, python held them as lists which tmk do not have a `.shape` attribute. The ixdat reader later turns those lists into arrays, and those arrays do have `.shape`.

I removed the inspector from this PR, see https://github.com/ixdat/ixdat/pull/207#discussion_r4061653369

src/ixdat/readers/asimov.py
FrederikLizakJohansen: The "dataset ID" belongs to asimov's older dataset and dataset-version system. A "project file" is one file stored in a project, and a "project-file bundle" groups related project files that ASIMOV handles together. Each project file or bundle has its own ID. Files and bundles will be kept separate for the beta-release of asimov, but will likely be merged for the major 1.0.0 release of asimov; where the parsed-backend will take the shape of ixdat objects.

src/ixdat/readers/asimov.py
FrederikLizakJohansen: You're right. Here: f9c871db8650e805625441db32aa4eef78fe00d1 I added `_get_object_type(cls)` and compare its result with the payload's type. There is one extra case: the general `Spectrum.read()` also accepts a spectrum series.

src/ixdat/readers/asimov.py
FrederikLizakJohansen: Good point. Here: b27cd5dcb4a696ea014abe4be6dbfe780f708cb8 I've added `object_class` to the docstring and explained that it tells the reader which ixdat class's declared fields it may copy from the payload.

src/ixdat/readers/asimov.py
FrederikLizakJohansen: Very very good point. My suggestion is that for the next release of ixdat, asimov will only be accessible through the asimov reader. The purpose of this first integration is simply to establish communication between the two software packages. Data from asimov can be loaded into ixdat, but this will be an intermediate solution. Before the official asimov 1.0.0 release, the asimov backend would then be redesigned around the ixdat backend model. This would be the point where the temporary reader/payload-based integration is replaced by the more direct backend architecture inside of ixdat.

tests/unit/test_asimov_reader.py
FrederikLizakJohansen: Agreed, we run into the risk of drift and with unit tests, that would break ixdat for no good reason. I trimmed the ixdat tests in 6216e6ffe3fe0054b2e2024e0de3d298905d456c to the reader's core behaviors. Since they use fixed examples, asimov should test that its exports still match the format ixdat reads. The demo provides a manual end-to-end check.

FrederikLizakJohansen:
Thank you for all your great comments @ScottSoren. I have tried to resolve all of them. You should hopefully be able to run the demo script now and any other custom IDs that you might have. Fingers crossed!

src/ixdat/readers/asimov.py
ScottSoren: In wonder if "would then be designed around the ixdat backend model" could become "would then _be_ the ixdat backend model" ... to avoid duplicate the code that e.g. keeps track of a measurement's data_series.
But we agree this is a problem for future versions of both packages :)

FrederikLizakJohansen: right, ideally yes! 🙌

src/ixdat/readers/asimov.py
ScottSoren: I thought it would remove more lines of code than it added :P ... oh well.

development_scripts/reader_demonstrators/inspect_asimov_payloads.py
ScottSoren: Issues described here were solved (thanks!), but new issues appear when running it, see high-level comment.

development_scripts/reader_demonstrators/demo_asimov_reader.py
```diff
@@ -1,47 +1,28 @@
...
 import matplotlib.pyplot as plt
 
-from ixdat import Measurement
+from ixdat import Measurement, Spectrum
 
-DATASET_IDS = [
-    "53ca3c5f-cd75-406c-8d99-a51c397988d7",
-    "32d304fa-9439-4e97-801e-26cc59c67d37",
-]
+NMR_SPECTRUM_ID = "300f4dcf-1585-51a6-81bc-f867910efe1a"
```
ScottSoren, line 11: I don't have access :(

development_scripts/reader_demonstrators/demo_asimov_reader.py
```diff
@@ -1,47 +1,28 @@
...
-    asimov_meta = (measurement.metadata or {}).get("asimov", {})
-    return (
-        asimov_meta.get("dataset_label")
-        or asimov_meta.get("dataset_id")
-        or measurement.name
-    )
+def load_demo_objects():
+    """Return the permanent NMR spectrum and Biologic CV Asimov examples."""
+    nmr_spectrum = Spectrum.read(NMR_SPECTRUM_ID, reader="asimov")
+    biologic_cv = Measurement.read(BIOLOGIC_CV_ID, reader="asimov")
```
ScottSoren, line 18: I get a problem here when trying to read a biologic file that I do have access to, e.g.
meas = Measurement.read("5fe83057-bc22-4c87-bf01-93954a2da305", reader="asimov")
```
ValueError: Asimov parsed payload response for 5fe83057-bc22-4c87-bf01-93954a2da305 has no payload_json or payload_uri.
```

development_scripts/reader_demonstrators/demo_asimov_reader.py
```diff
@@ -1,47 +1,28 @@
...
+    return nmr_spectrum, biologic_cv
 
 
 def main():
```
ScottSoren, line 22: As asked in the SEC PR: Is there a reason to protect the demo script functionality behind a main() function?
The convenience argument against it (in my workflow) is that a flat demo script when run in an ipython console leaves objects in namespace for further exploring.

ScottSoren (approved):
Excellent! It's beautiful how easy it is to find the id's and even copy reader code from asimov now.
I don't have access to the files in the demo script. Instead trying to demo with files I do have access to:
```
spec = Spectrum.read("01392fb0-2149-44a6-bfb5-351c5f4379ce", reader="asimov")
spec.plot()
```
Works beautifully. That's an asimov-uploaded xrdml file, which reads as a plain `Specrum`, as it should. It also gives a very legible error message if I try to use `Measurement` instead.

```
meas = Measurement.read("0c68d1e5-b26e-4789-adf8-d73b52127ede", reader="asimov")
meas.plot()
```
(an asimov-uploaded biologic file in the ixdat demo and tutorials project) similarly works fine as before :) 

GOOD TO MERGE from here!

development_scripts/reader_demonstrators/inspect_asimov_payloads.py
ScottSoren: Actually, it seems to work fine. I was just unfortunate enough to choose an asimov file that broke during upload, I think: 
```meas = Measurement.read("7516e34b-09a1-4b45-9ffb-f1abb145c882", reader="asimov")```
gives 
```
ValueError: Asimov parsed payload response for 7516e34b-09a1-4b45-9ffb-f1abb145c882 has no payload_json or payload_uri.
```
which is a bit of a cryptic error. But I guess if very few files on asimov are broken in that way, it's possibly not worth spending the time to have it diagnose and give an easier-to-understand error message.

File: development_scripts/reader_demonstrators/demo_asimov_reader.py (hunk 13 above)
Comment (FrederikLizakJohansen) [reply development_scripts/reader_demonstrators/demo_asimov_reader.py:22]: You can have a look at my thoughts in #212. The general gist is that with "main" we separate the "loading" from the "plotting" which means that we can reuse the loading functions in future scripts by simply importing (i.e. the demo scripts are doubling as modules). I also think a pattern like the one below enforces a standardization of the demo scripts that is really easy to navigate:
```python
  def load_example(data_dir=None):
      root = Path(data_dir) if data_dir is not None else get_demo_data_dir()
      return Measurement.read(root / "autolab/autolab_test_file.txt", reader="autolab")

  def main():
      load_example().plot()

  if __name__ == "__main__":
      main()
```

development_scripts/reader_demonstrators/demo_asimov_reader.py
FrederikLizakJohansen: I thought I made these files public; I will fix that asap.

development_scripts/reader_demonstrators/demo_asimov_reader.py
FrederikLizakJohansen: https://github.com/ixdat/ixdat/pull/207#discussion_r4069131667

development_scripts/reader_demonstrators/inspect_asimov_payloads.py
FrederikLizakJohansen: https://github.com/ixdat/ixdat/pull/207#discussion_r4069131667

PR #209, opened 2026-07-29
Add relational schema generation and SQLite storage backend

FrederikLizakJohansen:
## Summary

This PR lets ixdat save a complete project in one SQLite database file.

```python
from ixdat import Measurement
from ixdat.db import change_database

change_database("sqlite", db_path="my_project.sqlite")
measurement_id = measurement.save()

measurement = Measurement.get(measurement_id)
measurement = Measurement.load("my measurement")
```

SQLite is included with Python and adds zero database-package dependencies.

The database file can hold measurements, data series, spectra, calculators, calibrations, and the connections between them. Large NumPy arrays stay in the file until the user asks for their data.

## How the database layout is decided

ixdat classes already describe much of what they save:

- `table_name` says which table holds the main row.
- `column_attrs` lists the values stored in that row.
- `extra_column_attrs` lists extra values added by a more specialized class.
- `extra_linkers` lists other saved objects that belong to this object.

This PR adds two missing details:

- `column_types` says whether a value is text, a number, a dictionary, a list, a tuple, or a NumPy array.
- `column_references` says when an id points to one row in another table.

Each class describes only the information it adds. Information already described by a parent class is reused.

### How parent classes are combined

Some ixdat classes have more than one parent. `ECMSMeasurement`, for example, gets behavior and saved information from both `ECMeasurement` and `MSMeasurement`.

Python has an official order for walking through a class and its parents. This is called the **method resolution order**, usually shortened to **MRO**:

```text
ECMSMeasurement -> ECMeasurement -> MSMeasurement
                -> Measurement -> Saveable -> object
```

ixdat follows this same order when collecting information for the database. Each class in the list gets a chance to contribute the saved values it introduces. A parent that adds behavior without adding a saved value simply contributes no new column. Each actual contribution is collected once.

For an `ECMSMeasurement`:

- common values such as `name` and `technique` go in `measurement`;
- `ec_technique` goes in `ec_measurements`;
- `tspan_bg` goes in `ecms_measurements`.

All three rows use the same id, so SQLite knows that they describe one Python object.

Invalid type names and descriptions of values that do not exist raise a `DataBaseError` that names the class and value involved.

## How Python objects become tables

The new `ixdat.backends.relational` module turns the information above into a simple database blueprint (**schema**).

There are three main table shapes:

| Information in Python | Table shape | What one row means |
| --- | --- | --- |
| Values shared by one family of objects | Main table | One saved object |
| Extra values added by a specialized class | Extra table | Extra information for the same object id |
| An ordered list of connected objects | Connection table | One item and its place in the list |

A measurement produces a layout like this:

```text
measurement(
    id, name, aliases, metadata, sample_name, technique, tstamp
)

ec_measurements(
    id -> measurement.id,
    ec_technique
)

measurement_series(
    measurement_id -> measurement.id,
    position,
    data_series_id -> data_series.id
)
```

An arrow means that an id points to an existing row in another table. SQLite checks these pointers, which keeps objects from pointing to missing data.

The `position` value remembers the order of a Python list. A measurement can therefore load its data series in the same order in which they were saved. The same data series can also appear more than once in a list.

### Spectra use the same system

Spectral objects fit into the same three table shapes:

- A `Spectrum` points to one `Field` containing its numbers.
- A `Field` points to its axes in the correct order.
- A `SpectrumSeries` uses one axis for time and another for the scanned value, such as wavelength.
- A `MultiSpectrum` points to several fields in the correct order.
- A `SpectroMeasurement` points to one `SpectrumSeries`.
- An `ECOpticalMeasurement` can also point to a separate reference `Spectrum`.

SQLite makes the required tables when one of these objects is first saved. `SQLiteBackend.schema_report()` returns the exact table-building SQL for anyone who wants to inspect the full layout.

## Saving and loading

`SQLiteBackend` follows the same public interface as the existing ixdat backends:

- `save()` saves an object and everything it owns.
- `get(id)` loads one exact row.
- `load(name)` loads the newest row with that exact name.

Saving a complete group of connected objects is **all-or-nothing**. Database systems call this a transaction. If saving any part fails, SQLite removes all rows started by that save. This prevents half-saved measurements and loose child rows.

When ixdat loads a measurement, data series, or calculator, the saved kind tells ixdat which specialized Python class to rebuild. A loaded EC measurement therefore comes back with its EC behavior, and a loaded calibration comes back connected to its measurement.

NumPy arrays are stored inside the database file. Loading an object first reads its name, type, and connections. The larger numerical array is read when `.data` is accessed. This keeps the first load quick.

SQLite also checks every stored pointer and creates lookup guides for commonly searched values such as names and techniques. Database systems call these lookup guides indexes.

## Opening database files made by another ixdat version

The database file stores a layout version number.

When a future ixdat version adds a new optional value, ixdat can add an empty place for it to an older database automatically.

Changes that could alter the meaning of existing data receive a more careful response. Examples include changing text into a number or changing what an id points to. ixdat stops with a clear error so the database can be updated deliberately.

A file written with a newer layout version is also rejected. This prevents an older ixdat installation from guessing how to read a newer format.

## Relation to PR #75

[PR #75](https://github.com/ixdat/ixdat/pull/75) asked whether ixdat's Python classes could contain enough information to describe a relational database. The goal was to keep knowledge about saved values beside the classes that own those values.

A schema-generation experiment showed that the main idea worked. It also found several places where a database backend would still have to guess:

- What kind of value does each column contain?
- Which ids point to rows in another table?
- How should the two sides of a connection be named?
- How should list order be stored?
- How can every imported `Saveable` class be found?
- How should a class with several parents be handled?

This PR answers those questions and adds the SQLite backend listed as a follow-up in #75:

1. **One source for saved information.** Each `Saveable` class describes the values it owns. The SQLite backend reads that description, leaving the class information as the single source for the ixdat table layout.

2. **Familiar Python types.** Classes describe values with `int`, `float`, `str`, `dict`, `list`, `tuple`, and `numpy.ndarray`. SQLite decides how each Python type is written to the file.

3. **Clear connections between rows.** A class explicitly says when an id points to another saved object. Connection-table column names follow one simple rule based on the table names.

4. **List order is saved.** Each connection in a list receives a position. The object id and position identify that list entry.

5. **Imported classes are found automatically.** ixdat walks Python's existing list of `Saveable` child classes. Plugin classes join the same system as soon as they are imported.

6. **Classes with several parents follow Python's MRO.** Each parent contributes the saved values it introduced. All related rows share one object id.

This approach keeps the information close to the Python model. Inherited columns are stored once, and the SQLite backend reads the shared description.

## Small persistence fixes included here

Building and testing the complete database layout exposed several existing save-and-load gaps:

- Old or misspelled table references were corrected.
- `MultiSpectrum` now saves and reloads its fields and shared x-axis correctly.
- MS calibrations and background sets now reload their saved child objects only when those objects are requested.
- `ECOpticalMeasurement` now saves and reloads its reference spectrum.
- The directory backend can now write metadata containing NumPy values.
- Collecting all saved values no longer changes the class's original list.

These fixes have round-trip coverage with both the directory and SQLite backends.

## Validation

```text
python -m pytest tests/unit/test_sqlite_backend.py tests/functional/test_backends.py -q
# 44 passed

python -m pytest -q
# 228 passed; 4 existing Bruker reader setup errors also reproduce on main

invoke lint
invoke check-code-format
```

The tests cover:

- real and composed measurements;
- a basic `Spectrum`;
- an ordered `MultiSpectrum`;
- an EC-Optical measurement containing a `SpectrumSeries`;
- EC-Optical loading with and without a reference `Spectrum`;
- classes with several parents;
- calibrations and backgrounds;
- delayed loading of numerical arrays and connected child objects;
- all-or-nothing saving;
- loading by exact name;
- old and incompatible table layouts;
- several connections to one database file;
- temporary in-memory databases;
- checked pointers between tables.

The demo uses repository test data and a small made-up EC-Optical example:

```text
python development_scripts/demo_sqlite_backend.py
```

It shows an XRD `MultiSpectrum`, a time-resolved `SpectrumSeries`, its reference `Spectrum`, delayed array loading, generated tables, all-or-nothing saving, and closing and reopening the complete database.

## Review focus

The most useful areas to review are:

1. **Is each piece of information described in one place?** A class should list the new information that it saves. Information already described by a parent class should be reused. Please look for places where the same saved value is described more than once.

2. **Can ixdat save the kinds of values people normally use?** The current list covers whole numbers, decimal numbers, text, dictionaries, lists, tuples, and NumPy arrays. Please consider whether a typical ixdat plugin would need another common kind of value.

3. **Do saved objects stay connected correctly?** A measurement can point to data series, calculators, component measurements, and spectra. Some of these are ordered lists. Please check that saving and loading keeps every connection and keeps list items in the same order.

4. **What should happen when ixdat changes later?** If a future ixdat version adds a new optional value, it can add an empty place for that value to an old database automatically. If a change could alter the meaning of existing data, ixdat stops and explains that the database needs a careful manual update. Please check whether that is a sensible starting rule.

### DIFF
--- a/development_scripts/demo_sqlite_backend.py
+++ b/development_scripts/demo_sqlite_backend.py
@@ -230,8 +230,8 @@ print(
 print(f"  spectrum array before access: {lazy_optical_field._data}")
 print(f"  spectrum array after access:  shape={loaded_optical.spectra.data.shape}")
 
-# These are the table names that hold the pointers described above. They keep
-# relationships out of the main rows and allow lists to keep their order.
+# These tables hold the connections described above. The list relationships store a
+# position for each item. The single reference stores one spectrum id.
 print(
     "  relational paths: multispectrum_fields for XRD fields, field_axes for "
     "spectral axes, and ec_optical_measurements for the reference"
--- a/development_scripts/demo_sqlite_backend.py
+++ b/development_scripts/demo_sqlite_backend.py
@@ -170,6 +170,11 @@ part("8. The database speaks SQL: instant analytics over everything saved")
 # --------------------------------------------------------------------------- #
 
 print("All measurements, joined with their EC details and series counts:\n")
+# NOTE: this join only finds pure ECMeasurement rows, like the ones saved above.
+# An ECMSMeasurement writes its ec_technique to "ecms_measurements" instead of
+# "ec_measurements" - see the "Known limitation" section of
+# docs/source/diving_deeper/backend.rst - so a query meant to cover every EC-ish
+# measurement should filter on measurement.technique or UNION both tables.
 print(
     pd.read_sql_query(
         """
--- a/development_scripts/demo_sqlite_backend.py
+++ b/development_scripts/demo_sqlite_backend.py
@@ -237,3 +237,6 @@ print(f"\nDone. The database file is yours to explore: {DB_FILE}")
 
 if not plt.isinteractive():
     plt.show()
+
+con.close()
+DB.backend.close()
--- a/development_scripts/demo_sqlite_backend.py
+++ b/development_scripts/demo_sqlite_backend.py
@@ -1,24 +1,7 @@
-# -*- coding: utf-8 -*-
 """A tour of ixdat's SQLite database backend.
 
-This demo saves real measurements to a relational database in a single local
-file and shows off everything you gain by doing so:
-
-1.  saving is one line, and the whole schema is created for you
-2.  everything lives in one portable file with referential integrity
-3.  loading is lazy - metadata first, numerical data only on demand
-4.  objects come back by name or id, as the class they were saved as,
-    with their calculators (e.g. calibrations) intact
-5.  measurements built from other measurements reuse their existing data rows
-6.  every ixdat data family works: measurements, spectra, calculators, ...
-7.  the database speaks SQL: query it with pandas or any SQLite tool
-8.  rows can be updated in place
-9.  the full relational schema of ixdat's data model is generated from the
-    classes themselves (what PR #75 drew by hand: https://dbdiagram.io/d/601b...)
-
-It only uses data files shipped in the ixdat repository, so it runs anywhere:
-
-    python demo_sqlite_backend.py
+This demo saves measurements and spectra to a relational database in a single
+local file and walks through the main persistence features:
 """
 
 import shutil
@@ -27,36 +10,36 @@ import tempfile
 import time
 from pathlib import Path
 
+import numpy as np
 import pandas as pd
 from matplotlib import pyplot as plt
 
 from ixdat import Measurement, Spectrum
+from ixdat.data_series import DataSeries, Field, TimeSeries
 from ixdat.db import DB, change_database
+from ixdat.spectra import SpectrumSeries
+from ixdat.techniques.spectroelectrochemistry import ECOpticalMeasurement
 
 plt.close("all")
 
 TEST_DATA_DIR = Path(__file__).absolute().parent.parent / "test_data"
 
 DEMO_DIR = Path(tempfile.gettempdir()) / "ixdat_sqlite_demo"
-shutil.rmtree(DEMO_DIR, ignore_errors=True)  # start fresh on every run
+# Start with an empty folder so every run produces the same ids and row counts.
+shutil.rmtree(DEMO_DIR, ignore_errors=True)
 DB_FILE = DEMO_DIR / "demo_project.sqlite"
 
 
 def part(title):
-    print("\n" + "=" * 75 + f"\n{title}\n" + "=" * 75)
+    print("\n" + "=" * 75 + f"\n{title}\n" + "=" * 75 + "\n")
 
 
-# --------------------------------------------------------------------------- #
 part("1. One line to point ixdat at a database file")
-# --------------------------------------------------------------------------- #
 
 change_database("sqlite", db_path=DB_FILE)
 print(f"Active backend: {DB.backend}")
-print("(No new dependencies - this is python's built-in sqlite3.)")
 
-# --------------------------------------------------------------------------- #
 part("2. Saving a measurement (and its calibration) is one line each")
-# --------------------------------------------------------------------------- #
 
 ec = Measurement.read(TEST_DATA_DIR / "biologic/Pt_poly_cv_CUT.mpt", reader="biologic")
 ec.calibrate(RE_vs_RHE=0.72, A_el=0.196)  # a Calculator, saved along with it
@@ -64,18 +47,17 @@ ec.calibrate(RE_vs_RHE=0.72, A_el=0.196)  # a Calculator, saved along with it
 tic = time.perf_counter()
 ec_id = ec.save()
 toc = time.perf_counter()
-print(f"Saved {ec!r} with all its data series in {(toc - tic) * 1000:.1f} ms")
+print(
+    f"Saved {ec!r} in {DB_FILE} with all its data series in {(toc - tic) * 1000:.1f} ms"
+)
+print(f"  ({DB_FILE.stat().st_size / 1024:.0f} kB)")
 
-# --------------------------------------------------------------------------- #
-part("3. What that created: a real relational database in a single file")
-# --------------------------------------------------------------------------- #
-
-print(f"Everything is in the one file {DB_FILE}")
-print(f"  ({DB_FILE.stat().st_size / 1024:.0f} kB - open it with any SQLite tool,")
-print("   e.g. DB Browser for SQLite, DBeaver, datasette, or the sqlite3 CLI)\n")
-
-# The tables were created on demand, derived from the ixdat classes involved:
-con = sqlite3.connect(DB_FILE)  # a plain SQL connection, to peek behind the scenes
+# A database table is much like one sheet in a spreadsheet. Each row holds one
+# saved item. ixdat reads the class definitions and makes the needed sheets when
+# an item is first saved. The table plan stays beside the Python code.
+#
+# sqlite3 is Python's built-in way to look directly inside the database file.
+con = sqlite3.connect(DB_FILE)
 tables = [
     row[0]
     for row in con.execute(
@@ -86,6 +68,9 @@ for table in tables:
     count = con.execute(f'SELECT COUNT(*) FROM "{table}"').fetchone()[0]
     print(f"  table {table:.<28} {count:>3} rows")
 
+# A foreign key is an id that points to a row in another table. For example, a
+# measurement stores the ids of its data series. SQLite checks these pointers,
+# which prevents a measurement from pointing to a row that does not exist.
 print("\nWith honest-to-goodness foreign keys, e.g.:\n")
 print(
     con.execute(
@@ -95,10 +80,26 @@ print(
 violations = con.execute("PRAGMA foreign_key_check").fetchall()
 print(f"\nForeign key violations in the database: {violations or 'none'}")
 
-# --------------------------------------------------------------------------- #
-part("4. Loading is lazy: metadata now, numbers only when you need them")
-# --------------------------------------------------------------------------- #
+# The schema version says which table layout the file uses. ixdat checks it
+# before reading the file. An index is a small lookup guide that helps SQLite
+# find common values, such as a name or technique, without scanning every row.
+schema_version = con.execute(
+    'SELECT value FROM "_ixdat_metadata" WHERE key = ?', ("schema_version",)
+).fetchone()[0]
+lookup_indexes = con.execute(
+    """
+    SELECT COUNT(*) FROM sqlite_master
+    WHERE type = 'index' AND name LIKE 'ixdat_%'
+    """
+).fetchone()[0]
+print(
+    f"Storage schema version: {schema_version}; lookup indexes created: {lookup_indexes}"
+)
 
+part("3. Loading is lazy: metadata now, numbers only when you need them")
+
+# Names and other small details are cheap to load. Numerical arrays can be much
+# larger, so ixdat leaves them in the database until code asks for `.data`.
 tic = time.perf_counter()
 loaded_ec = Measurement.get(ec_id)
 toc = time.perf_counter()
@@ -115,10 +116,10 @@ print(
     f"in {(toc - tic) * 1000:.2f} ms"
 )
 
-# --------------------------------------------------------------------------- #
-part("5. Objects come back by name, as what they were, calibration included")
-# --------------------------------------------------------------------------- #
+part("4. Objects come back by name, as what they were, calibration included")
 
+# The saved technique tells ixdat which kind of Measurement to rebuild. Links
+# stored in the database also reconnect the measurement to its calibration.
 by_name = Measurement.load("Pt_poly_cv_CUT.mpt")
 print(f"Measurement.load('Pt_poly_cv_CUT.mpt') -> {by_name!r}")
 print(f"  concrete class: {type(by_name).__name__}")
@@ -132,14 +133,15 @@ print(
     "  <- the RE_vs_RHE=0.72 calibration survived the round trip"
 )
 
-# --------------------------------------------------------------------------- #
-part("6. Composed measurements reuse their component data rows")
-# --------------------------------------------------------------------------- #
+part("5. Composed measurements reuse their component data rows")
 
 composed = ec.select(cycle=1) + ec.select(cycle=3)
 composed_id = composed.save()
 print(f"Saved composed measurement (id={composed_id}) built from two cycles.\n")
 
+# The measurement_series table is a list of pointers from measurements to data
+# series. Several measurements can point to the same series row. Large arrays
+# are then stored once even when they are used in several places.
 n_references, n_series = con.execute(
     "SELECT COUNT(*), COUNT(DISTINCT data_series_id) FROM measurement_series"
 ).fetchone()
@@ -151,24 +153,117 @@ print("  the composed measurement and its components *share* rows.")
 reloaded_composed = Measurement.get(composed_id)
 print(f"\nRound trip of the composed measurement: {reloaded_composed == composed}")
 
-# --------------------------------------------------------------------------- #
-part("7. Spectra and other ixdat data families use the same backend")
-# --------------------------------------------------------------------------- #
+part("6. Spectra and other ixdat data families use the same backend")
 
+# A MultiSpectrum groups several spectral fields that share one x-axis. The
+# database keeps pointers to those fields in the same order as the Python list.
 xrd = Spectrum.read(TEST_DATA_DIR / "xrd/twotheta_2th.xy", reader="xrdxy")
 xrd_id = xrd.save()
 loaded_xrd = type(xrd).get(xrd_id)
-print(f"Saved and reloaded {loaded_xrd!r}")
-print(f"  class: {type(loaded_xrd).__name__}, equal: {loaded_xrd == xrd}")
 print(
-    "  (spectra live in their own tables: multispectrum, data_series, "
-    "field_axes, ...)"
+    f"XRD: {type(loaded_xrd).__name__}, {len(loaded_xrd.fields)} field(s), "
+    f"equal after reload: {loaded_xrd == xrd}"
 )
 
-# --------------------------------------------------------------------------- #
-part("8. The database speaks SQL: instant analytics over everything saved")
-# --------------------------------------------------------------------------- #
+# A SpectrumSeries is a grid of numbers. One axis is time and the other is
+# wavelength. The Field holds the grid and pointers to those two axes.
+wavelength = DataSeries(
+    name="wavelength / nm",
+    unit_name="nm",
+    data=np.linspace(400, 700, 4),
+)
+optical_time = TimeSeries(
+    name="optical time / s",
+    unit_name="s",
+    data=np.array([0.0, 1.0, 2.0]),
+    tstamp=1.6e9,
+)
+optical_series = SpectrumSeries(
+    name="time-resolved optical spectra",
+    technique="EC-Optical",
+    tstamp=optical_time.tstamp,
+    field=Field(
+        name="intensity",
+        unit_name="counts",
+        data=np.array(
+            [
+                [1.0, 2.0, 3.0, 4.0],
+                [2.0, 3.0, 4.0, 5.0],
+                [3.0, 4.0, 5.0, 6.0],
+            ]
+        ),
+        axes_series=[optical_time, wavelength],
+    ),
+)
+reference = Spectrum(
+    name="optical reference",
+    technique="optical",
+    tstamp=optical_time.tstamp,
+    field=Field(
+        name="reference intensity",
+        unit_name="counts",
+        data=np.ones(4),
+        axes_series=[wavelength],
+    ),
+)
 
+# This measurement owns the changing SpectrumSeries above and points to one
+# separate Spectrum used as its reference. The database saves the whole group
+# and remembers how its pieces connect.
+optical = ECOpticalMeasurement(
+    name="synthetic EC-Optical measurement",
+    technique="EC-Optical",
+    ec_technique="synthetic potential sweep",
+    tstamp=optical_time.tstamp,
+    series_list=[optical_time],
+    spectrum_series=optical_series,
+    reference_spectrum=reference,
+)
+optical_id = optical.save()
+loaded_optical = Measurement.get(optical_id)
+lazy_optical_field = loaded_optical.spectrum_series.field
+print(
+    f"EC-Optical: {type(loaded_optical).__name__}, "
+    f"linked {type(loaded_optical.spectrum_series).__name__} and "
+    f"{type(loaded_optical.reference_spectrum).__name__}"
+)
+print(f"  spectrum array before access: {lazy_optical_field._data}")
+print(f"  spectrum array after access:  shape={loaded_optical.spectra.data.shape}")
+
+# These are the table names that hold the pointers described above. They keep
+# relationships out of the main rows and allow lists to keep their order.
+print(
+    "  relational paths: multispectrum_fields for XRD fields, field_axes for "
+    "spectral axes, and ec_optical_measurements for the reference"
+)
+
+part("7. A complete object graph is one transaction")
+
+# A transaction makes a save all-or-nothing. This example has metadata that
+# cannot be saved as JSON. SQLite removes the child series it had started to
+# save, so no loose or half-saved rows remain.
+rows_before = con.execute('SELECT COUNT(*) FROM "data_series"').fetchone()[0]
+invalid = Measurement(
+    name="this graph will roll back",
+    technique="simple",
+    metadata={"not JSON": object()},
+    series_list=[DataSeries(name="unsaved child", unit_name="V", data=np.array([1.0]))],
+)
+try:
+    invalid.save()
+except TypeError:
+    pass
+rows_after = con.execute('SELECT COUNT(*) FROM "data_series"').fetchone()[0]
+print(
+    "An invalid parent rolled back itself and its child: "
+    f"data-series rows stayed at {rows_before} ({rows_after == rows_before})."
+)
+
+part("8. The database speaks SQL: instant analytics over everything saved")
+
+# JOIN lines up rows from different tables using their ids. LEFT JOIN keeps a
+# measurement in the result even when it has no matching EC detail or series.
+# GROUP BY gathers all series links for one measurement so COUNT can total them.
 print("All measurements, joined with their EC details and series counts:\n")
 print(
     pd.read_sql_query(
@@ -197,10 +292,9 @@ print(
 )
 print("\n(Anything - pandas, R, LibreOffice, grafana - can query this file.)")
 
-# --------------------------------------------------------------------------- #
 part("9. Rows can be updated in place")
-# --------------------------------------------------------------------------- #
 
+# `force=True` writes the changed values back to the existing row with this id.
 loaded_ec.name = "Pt_poly_cv_CUT.mpt (analyzed 2026-07-02)"
 DB.backend.save(loaded_ec, force=True)
 print(f"Renamed and updated row id={loaded_ec.id}. The database now says:")
@@ -209,10 +303,11 @@ updated_name = con.execute(
 ).fetchone()[0]
 print(f"  {updated_name}")
 
-# --------------------------------------------------------------------------- #
 part("10. The whole schema of ixdat's data model, generated from the classes")
-# --------------------------------------------------------------------------- #
 
+# A schema is the database's blueprint: its tables, columns, and links. ixdat
+# builds this blueprint from the save information already written on its Python
+# classes. A plugin class can describe its own saved values in the same way.
 report = DB.backend.schema_report()
 schema_file = DEMO_DIR / "ixdat_schema.sql"
 schema_file.write_text(report)
@@ -220,24 +315,45 @@ n_tables = report.count("CREATE TABLE")
 print(f"schema_report() derived {n_tables} tables from the imported ixdat classes")
 print(f"  (full DDL written to {schema_file})")
 print("\nA taste - inheritance becomes extension tables joined on id:\n")
+
+# In Python, a TimeSeries is a DataSeries with an extra timestamp. SQL stores
+# the shared DataSeries values in one table and the timestamp in a small extra
+# table. Both rows use the same id, so SQLite knows they describe one object.
+# EC-MS measurements follow the same pattern for their extra values.
 for statement in report.split(";"):
     if '"ecms_measurements"' in statement or '"tstamps"' in statement:
         print(statement.strip() + ";\n")
 
-# --------------------------------------------------------------------------- #
-part("Finally: plots, straight from the database")
-# --------------------------------------------------------------------------- #
+part("11. Close and reopen the complete project")
 
-axes = Measurement.load("Pt_poly_cv_CUT.mpt (analyzed 2026-07-02)").plot_measurement()
+# Closing ends the current connection to the file. The saved rows stay on disk.
+# Loading from the same file rebuilds each requested object and its links.
+con.close()
+DB.backend.close()
+change_database("sqlite", db_path=DB_FILE)
+
+reopened_ec = Measurement.load("Pt_poly_cv_CUT.mpt (analyzed 2026-07-02)")
+reopened_xrd = type(xrd).get(xrd_id)
+reopened_optical = Measurement.get(optical_id)
+print(
+    f"Reopened {type(reopened_ec).__name__}, {type(reopened_xrd).__name__}, "
+    f"and {type(reopened_optical).__name__} from {DB_FILE.name}"
+)
+
+part("Finally: plots, straight from the reopened database")
+
+axes = reopened_ec.plot_measurement()
 axes[0].set_title("EC measurement reloaded from SQLite")
 
-ax = type(xrd).get(xrd_id).plot()
+ax = reopened_xrd.plot()
 ax.set_title("XRD spectrum reloaded from SQLite")
+
+ax = reopened_optical.spectrum_series.heat_plot()
+ax.set_title("SpectrumSeries reloaded through an EC-Optical measurement")
 
 print(f"\nDone. The database file is yours to explore: {DB_FILE}")
 
 if not plt.isinteractive():
     plt.show()
 
-con.close()
 DB.backend.close()
--- a/development_scripts/demo_sqlite_backend.py
+++ b/development_scripts/demo_sqlite_backend.py
@@ -9,7 +9,7 @@ file and shows off everything you gain by doing so:
 3.  loading is lazy - metadata first, numerical data only on demand
 4.  objects come back by name or id, as the class they were saved as,
     with their calculators (e.g. calibrations) intact
-5.  measurements built from other measurements share rows - no data duplication
+5.  measurements built from other measurements reuse their existing data rows
 6.  every ixdat data family works: measurements, spectra, calculators, ...
 7.  the database speaks SQL: query it with pandas or any SQLite tool
 8.  rows can be updated in place
@@ -121,7 +121,7 @@ part("5. Objects come back by name, as what they were, calibration included")
 
 by_name = Measurement.load("Pt_poly_cv_CUT.mpt")
 print(f"Measurement.load('Pt_poly_cv_CUT.mpt') -> {by_name!r}")
-print(f"  class: {type(by_name).__name__} (not just a generic Measurement)")
+print(f"  concrete class: {type(by_name).__name__}")
 print(f"  calculators: {by_name.calculator_list}")
 print(f"  equal to what was saved:  by_name == ec  is  {by_name == ec}")
 
@@ -133,7 +133,7 @@ print(
 )
 
 # --------------------------------------------------------------------------- #
-part("6. Composed measurements share rows - data is never duplicated")
+part("6. Composed measurements reuse their component data rows")
 # --------------------------------------------------------------------------- #
 
 composed = ec.select(cycle=1) + ec.select(cycle=3)
@@ -144,17 +144,15 @@ n_references, n_series = con.execute(
     "SELECT COUNT(*), COUNT(DISTINCT data_series_id) FROM measurement_series"
 ).fetchone()
 n_measurements = con.execute("SELECT COUNT(*) FROM measurement").fetchone()[0]
-print(f"  {n_measurements} measurements now reference series "
-      f"{n_references} times ...")
+print(f"  {n_measurements} measurements now reference series {n_references} times ...")
 print(f"  ... but only {n_series} distinct data series (arrays) are stored:")
 print("  the composed measurement and its components *share* rows.")
 
 reloaded_composed = Measurement.get(composed_id)
-print(f"\nRound trip of the composed measurement: "
-      f"{reloaded_composed == composed}")
+print(f"\nRound trip of the composed measurement: {reloaded_composed == composed}")
 
 # --------------------------------------------------------------------------- #
-part("7. Not just measurements: every ixdat data family gets its tables")
+part("7. Spectra and other ixdat data families use the same backend")
 # --------------------------------------------------------------------------- #
 
 xrd = Spectrum.read(TEST_DATA_DIR / "xrd/twotheta_2th.xy", reader="xrdxy")
@@ -162,19 +160,16 @@ xrd_id = xrd.save()
 loaded_xrd = type(xrd).get(xrd_id)
 print(f"Saved and reloaded {loaded_xrd!r}")
 print(f"  class: {type(loaded_xrd).__name__}, equal: {loaded_xrd == xrd}")
-print("  (spectra live in their own tables: multispectrum, data_series, "
-      "field_axes, ...)")
+print(
+    "  (spectra live in their own tables: multispectrum, data_series, "
+    "field_axes, ...)"
+)
 
 # --------------------------------------------------------------------------- #
 part("8. The database speaks SQL: instant analytics over everything saved")
 # --------------------------------------------------------------------------- #
 
 print("All measurements, joined with their EC details and series counts:\n")
-# NOTE: this join only finds pure ECMeasurement rows, like the ones saved above.
-# An ECMSMeasurement writes its ec_technique to "ecms_measurements" instead of
-# "ec_measurements" - see the "Known limitation" section of
-# docs/source/diving_deeper/backend.rst - so a query meant to cover every EC-ish
-# measurement should filter on measurement.technique or UNION both tables.
 print(
     pd.read_sql_query(
         """
@@ -209,9 +204,10 @@ part("9. Rows can be updated in place")
 loaded_ec.name = "Pt_poly_cv_CUT.mpt (analyzed 2026-07-02)"
 DB.backend.save(loaded_ec, force=True)
 print(f"Renamed and updated row id={loaded_ec.id}. The database now says:")
-print("  " + str(con.execute(
+updated_name = con.execute(
     "SELECT name FROM measurement WHERE id = ?", (loaded_ec.id,)
-).fetchone()[0]))
+).fetchone()[0]
+print(f"  {updated_name}")
 
 # --------------------------------------------------------------------------- #
 part("10. The whole schema of ixdat's data model, generated from the classes")
--- a/development_scripts/demo_sqlite_backend.py
+++ b/development_scripts/demo_sqlite_backend.py
@@ -1,0 +1,239 @@
+# -*- coding: utf-8 -*-
+"""A tour of ixdat's SQLite database backend.
+
+This demo saves real measurements to a relational database in a single local
+file and shows off everything you gain by doing so:
+
+1.  saving is one line, and the whole schema is created for you
+2.  everything lives in one portable file with referential integrity
+3.  loading is lazy - metadata first, numerical data only on demand
+4.  objects come back by name or id, as the class they were saved as,
+    with their calculators (e.g. calibrations) intact
+5.  measurements built from other measurements share rows - no data duplication
+6.  every ixdat data family works: measurements, spectra, calculators, ...
+7.  the database speaks SQL: query it with pandas or any SQLite tool
+8.  rows can be updated in place
+9.  the full relational schema of ixdat's data model is generated from the
+    classes themselves (what PR #75 drew by hand: https://dbdiagram.io/d/601b...)
+
+It only uses data files shipped in the ixdat repository, so it runs anywhere:
+
+    python demo_sqlite_backend.py
+"""
+
+import shutil
+import sqlite3
+import tempfile
+import time
+from pathlib import Path
+
+import pandas as pd
+from matplotlib import pyplot as plt
+
+from ixdat import Measurement, Spectrum
+from ixdat.db import DB, change_database
+
+plt.close("all")
+
+TEST_DATA_DIR = Path(__file__).absolute().parent.parent / "test_data"
+
+DEMO_DIR = Path(tempfile.gettempdir()) / "ixdat_sqlite_demo"
+shutil.rmtree(DEMO_DIR, ignore_errors=True)  # start fresh on every run
+DB_FILE = DEMO_DIR / "demo_project.sqlite"
+
+
+def part(title):
+    print("\n" + "=" * 75 + f"\n{title}\n" + "=" * 75)
+
+
+# --------------------------------------------------------------------------- #
+part("1. One line to point ixdat at a database file")
+# --------------------------------------------------------------------------- #
+
+change_database("sqlite", db_path=DB_FILE)
+print(f"Active backend: {DB.backend}")
+print("(No new dependencies - this is python's built-in sqlite3.)")
+
+# --------------------------------------------------------------------------- #
+part("2. Saving a measurement (and its calibration) is one line each")
+# --------------------------------------------------------------------------- #
+
+ec = Measurement.read(TEST_DATA_DIR / "biologic/Pt_poly_cv_CUT.mpt", reader="biologic")
+ec.calibrate(RE_vs_RHE=0.72, A_el=0.196)  # a Calculator, saved along with it
+
+tic = time.perf_counter()
+ec_id = ec.save()
+toc = time.perf_counter()
+print(f"Saved {ec!r} with all its data series in {(toc - tic) * 1000:.1f} ms")
+
+# --------------------------------------------------------------------------- #
+part("3. What that created: a real relational database in a single file")
+# --------------------------------------------------------------------------- #
+
+print(f"Everything is in the one file {DB_FILE}")
+print(f"  ({DB_FILE.stat().st_size / 1024:.0f} kB - open it with any SQLite tool,")
+print("   e.g. DB Browser for SQLite, DBeaver, datasette, or the sqlite3 CLI)\n")
+
+# The tables were created on demand, derived from the ixdat classes involved:
+con = sqlite3.connect(DB_FILE)  # a plain SQL connection, to peek behind the scenes
+tables = [
+    row[0]
+    for row in con.execute(
+        "SELECT name FROM sqlite_master WHERE type='table' ORDER BY name"
+    )
+]
+for table in tables:
+    count = con.execute(f'SELECT COUNT(*) FROM "{table}"').fetchone()[0]
+    print(f"  table {table:.<28} {count:>3} rows")
+
+print("\nWith honest-to-goodness foreign keys, e.g.:\n")
+print(
+    con.execute(
+        "SELECT sql FROM sqlite_master WHERE name='measurement_series'"
+    ).fetchone()[0]
+)
+violations = con.execute("PRAGMA foreign_key_check").fetchall()
+print(f"\nForeign key violations in the database: {violations or 'none'}")
+
+# --------------------------------------------------------------------------- #
+part("4. Loading is lazy: metadata now, numbers only when you need them")
+# --------------------------------------------------------------------------- #
+
+tic = time.perf_counter()
+loaded_ec = Measurement.get(ec_id)
+toc = time.perf_counter()
+print(f"Measurement.get({ec_id}) took {(toc - tic) * 1000:.2f} ms")
+print("  ... because it read *no* numerical data:")
+series = loaded_ec.series_list[0]
+print(f"  {series.name!r}._data is {series._data}")
+
+tic = time.perf_counter()
+n_points = len(series.data)  # NOW the array is fetched from the database
+toc = time.perf_counter()
+print(
+    f"  first access to .data fetched {n_points} points "
+    f"in {(toc - tic) * 1000:.2f} ms"
+)
+
+# --------------------------------------------------------------------------- #
+part("5. Objects come back by name, as what they were, calibration included")
+# --------------------------------------------------------------------------- #
+
+by_name = Measurement.load("Pt_poly_cv_CUT.mpt")
+print(f"Measurement.load('Pt_poly_cv_CUT.mpt') -> {by_name!r}")
+print(f"  class: {type(by_name).__name__} (not just a generic Measurement)")
+print(f"  calculators: {by_name.calculator_list}")
+print(f"  equal to what was saved:  by_name == ec  is  {by_name == ec}")
+
+t, v_raw = by_name.grab("raw_potential")
+t, v = by_name.grab("potential")  # calibrated, thanks to the reloaded ECCalibration
+print(
+    f"  mean(potential - raw_potential) = {(v - v_raw).mean():.2f} V"
+    "  <- the RE_vs_RHE=0.72 calibration survived the round trip"
+)
+
+# --------------------------------------------------------------------------- #
+part("6. Composed measurements share rows - data is never duplicated")
+# --------------------------------------------------------------------------- #
+
+composed = ec.select(cycle=1) + ec.select(cycle=3)
+composed_id = composed.save()
+print(f"Saved composed measurement (id={composed_id}) built from two cycles.\n")
+
+n_references, n_series = con.execute(
+    "SELECT COUNT(*), COUNT(DISTINCT data_series_id) FROM measurement_series"
+).fetchone()
+n_measurements = con.execute("SELECT COUNT(*) FROM measurement").fetchone()[0]
+print(f"  {n_measurements} measurements now reference series "
+      f"{n_references} times ...")
+print(f"  ... but only {n_series} distinct data series (arrays) are stored:")
+print("  the composed measurement and its components *share* rows.")
+
+reloaded_composed = Measurement.get(composed_id)
+print(f"\nRound trip of the composed measurement: "
+      f"{reloaded_composed == composed}")
+
+# --------------------------------------------------------------------------- #
+part("7. Not just measurements: every ixdat data family gets its tables")
+# --------------------------------------------------------------------------- #
+
+xrd = Spectrum.read(TEST_DATA_DIR / "xrd/twotheta_2th.xy", reader="xrdxy")
+xrd_id = xrd.save()
+loaded_xrd = type(xrd).get(xrd_id)
+print(f"Saved and reloaded {loaded_xrd!r}")
+print(f"  class: {type(loaded_xrd).__name__}, equal: {loaded_xrd == xrd}")
+print("  (spectra live in their own tables: multispectrum, data_series, "
+      "field_axes, ...)")
+
+# --------------------------------------------------------------------------- #
+part("8. The database speaks SQL: instant analytics over everything saved")
+# --------------------------------------------------------------------------- #
+
+print("All measurements, joined with their EC details and series counts:\n")
+print(
+    pd.read_sql_query(
+        """
+        SELECT m.id, m.name, m.technique, e.ec_technique,
+               COUNT(ms.data_series_id) AS n_series
+        FROM measurement m
+        LEFT JOIN ec_measurements e ON e.id = m.id
+        LEFT JOIN measurement_series ms ON ms.measurement_id = m.id
+        GROUP BY m.id ORDER BY m.id
+        """,
+        con,
+    ).to_string(index=False)
+)
+
+print("\nAll calculators, joined with their calibration constants:\n")
+print(
+    pd.read_sql_query(
+        """
+        SELECT c.id, c.name, c.calculator_type, cal.RE_vs_RHE, cal.A_el, cal.R_Ohm
+        FROM calculator c
+        LEFT JOIN ec_calibration cal ON cal.id = c.id
+        """,
+        con,
+    ).to_string(index=False)
+)
+print("\n(Anything - pandas, R, LibreOffice, grafana - can query this file.)")
+
+# --------------------------------------------------------------------------- #
+part("9. Rows can be updated in place")
+# --------------------------------------------------------------------------- #
+
+loaded_ec.name = "Pt_poly_cv_CUT.mpt (analyzed 2026-07-02)"
+DB.backend.save(loaded_ec, force=True)
+print(f"Renamed and updated row id={loaded_ec.id}. The database now says:")
+print("  " + str(con.execute(
+    "SELECT name FROM measurement WHERE id = ?", (loaded_ec.id,)
+).fetchone()[0]))
+
+# --------------------------------------------------------------------------- #
+part("10. The whole schema of ixdat's data model, generated from the classes")
+# --------------------------------------------------------------------------- #
+
+report = DB.backend.schema_report()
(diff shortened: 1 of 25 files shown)


development_scripts/demo_sqlite_backend.py
```diff
@@ -0,0 +1,359 @@
...
+        con,
+    ).to_string(index=False)
+)
+print("\n(Anything - pandas, R, LibreOffice, grafana - can query this file.)")
+
+part("9. Rows can be updated in place")
+
+# `force=True` writes the changed values back to the existing row with this id.
+loaded_ec.name = "Pt_poly_cv_CUT.mpt (analyzed 2026-07-02)"
+DB.backend.save(loaded_ec, force=True)
```
ScottSoren, line 299: does this re-save the data series or just confirm that they're already there?

development_scripts/demo_sqlite_backend.py
```diff
@@ -0,0 +1,359 @@
...
+# EC-MS measurements follow the same pattern for their extra values.
+for statement in report.split(";"):
+    if '"ecms_measurements"' in statement or '"tstamps"' in statement:
+        print(statement.strip() + ";\n")
+
+part("11. Close and reopen the complete project")
+
+# Closing ends the current connection to the file. The saved rows stay on disk.
+# Loading from the same file rebuilds each requested object and its links.
+con.close()
```
ScottSoren, line 331: are `con` and `DB.backend` two names for the same thing?

development_scripts/demo_sqlite_backend.py
```diff
@@ -0,0 +1,359 @@
...
+DB.backend.close()
+change_database("sqlite", db_path=DB_FILE)
```
ScottSoren, line 333: This is identical to line 39. I guess it reopens the connection. Something a bit un-intuitive about that behavior, but makes sense.

development_scripts/demo_sqlite_webapp.py
```diff
@@ -0,0 +1,92 @@
...
+DATABASE = Path(tempfile.gettempdir()) / "ixdat_web_demo.sqlite"
+
+
+def save_data():
+    if DATABASE.exists():
+        DATABASE.unlink()
+    change_database("sqlite", db_path=DATABASE)
+    Measurement.read(DATA / "biologic/Pt_poly_cv_CUT.mpt", reader="biologic").save()
+    Measurement.read(DATA / "biologic/Pt_poly_cv.mpt", reader="biologic").save()
+    Spectrum.read(DATA / "xrd/twotheta_2th.xy", reader="xrdxy").save()
```
ScottSoren, line 25: I tried also saving a SpectrumSeries here and it works :) 

in axes_series you get a data_series_id appearing on two rows with different linked_data_series_id's, as you should.

development_scripts/demo_sqlite_webapp.py
```diff
@@ -0,0 +1,92 @@
...
+    DB.backend.close()
+
+
+def make_page(selected):
+    db = sqlite3.connect(DATABASE.resolve().as_uri() + "?mode=ro", uri=True)
+    tables = [
```
ScottSoren, line 31: looks like this query gets made every time the viewer changes page, whereas it could be called just once?

development_scripts/demo_sqlite_webapp.py
```diff
@@ -0,0 +1,92 @@
...
+    links = " ".join(
+        '<a href="/?table={0}">{0}</a>'.format(html.escape(name)) for name in tables
+    )
+    headings = "".join("<th>{}</th>".format(html.escape(name)) for name in columns)
+    body = "".join(
+        "<tr>{}</tr>".format("".join("<td>{}</td>".format(show(value)) for value in row))
+        for row in rows
+    )
+    return (
+        "<!doctype html><title>ixdat SQLite tables</title>"
```
ScottSoren, line 53: wow

src/ixdat/backends/memory_backend.py
```diff
@@ -30,6 +31,17 @@ def get(self, cls, i):
         """Return an object of a specified class and id by looking up in memory"""
         return self.objects[cls.table_name][i]
 
+    def load(self, cls, name):
+        """Return the most recently saved object of cls with the given name"""
+        objects_by_id = self.objects.get(cls.table_name, {})
```
ScottSoren, line 36: does this select all of the objects, with all of their columns? Wouldn't it be enough just to select the id and name columns, figure out which object to return, and then select that one object?
(Effectively, that's what you do in `DirBackend.load`)

docs/source/diving_deeper/backend.rst
```diff
@@ -1,6 +1,132 @@
 .. _backend:
 
-Backend
-==============================================
+Database backends
```
ScottSoren, line 3: Yes! :D 
Thanks for putting stuff directly in docs.

src/ixdat/backends/relational.py
```diff
@@ -0,0 +1,283 @@
...
+
+        Args:
+            name (str): The name of the table
+            columns (list of ColumnSchema): The columns, starting with "id"
+            extends (str or None): The name of the main table this table
+                extends, if it is an extension table
+        """
+        self.name = name
+        self.columns = columns
+        self.extends = extends
```
ScottSoren: is `extends` an established SQL concept?

src/ixdat/backends/relational.py
```diff
@@ -0,0 +1,283 @@
...
+            collect(subclass)
+
+    collect(Saveable)
+    return classes
+
+
+def family_table_schemas(cls):
+    """Return the extension and linker tables of all classes sharing cls's main table
+
+    Before reading a row, its exact subclass is unknown (a row in "measurement"
```
ScottSoren, line 263: I think this isn't actually necessary with the `technique` column. If an entry in the measurement table has technique="EC-MS", then it must also have a row in the ecms_measurement table.

src/ixdat/backends/relational.py
```diff
@@ -0,0 +1,283 @@
...
+See :ref:`backend` for the persistence model, supported types, inheritance,
+and linked objects.
+"""
+
+import numpy as np
+
+from ..exceptions import DataBaseError
+
+
+KNOWN_COLUMN_TYPES = (int, float, str, dict, list, tuple, np.ndarray)
```
ScottSoren, line 16: I'm in doubt about whether ixdat also uses sets in column attributes.

src/ixdat/techniques/spectroelectrochemistry.py
```diff
@@ -68,6 +68,9 @@ def __init__(self, reference_spectrum=None, ref_id=None, **kwargs):
             self._reference_spectrum = reference_spectrum
         elif ref_id:
             self._reference_spectrum = PlaceHolderObject(ref_id, cls=Spectrum)
+        else:
+            # a reference spectrum can also be set later, with set_reference_spectrum:
+            self._reference_spectrum = None
```
ScottSoren, line 76: I think Olly also did this on her branch...

src/ixdat/techniques/spectroelectrochemistry.py
```diff
@@ -81,6 +84,21 @@ def reference_spectrum(self):
             self._reference_spectrum = self._reference_spectrum.get_object()
         return self._reference_spectrum
 
+    # FIXME: The attribute below is needed in order to correctly pass the reference
+    #   spectrum between objects using its id, because "child_attrs" only works on
```
ScottSoren: Hm... but if it's not a many-to-many relationship, does it need to be a child_attr? Why not this?
```
extra_column_attrs = {
        "ec_optical_measurement": {
            "reference_spectrum_id",
        }
}
...
@property 
def reference_spectrum_id(self):
    return self.reference_spectrum.id
```

ScottSoren: I just realized the answer: that wouldn't guarantee that the reference spectrum is saved when the ECOpticalMeasurement is saved. Okay, yeah, that's a design flaw. 
I wonder, how does sqlalchemy make sure the rows referenced by foreign keys, whether one-to-one or many-to-many, are in the relevant tables when a row with foreign keys is added to a table? Is it something like child_attrs but more flexible, or is it completely different?

src/ixdat/techniques/ms.py
```diff
@@ -321,7 +321,6 @@ class MSSpectrumSeries(SpectrumSeries):
 
 
 class MSSpectroMeasurement(MSMeasurement, SpectroMeasurement):
-    extra_column_attrs = SpectroMeasurement.extra_column_attrs
```
ScottSoren, line 324: out of curiosity, why delete this line?

src/ixdat/calculators/ms_calculators.py
```diff
@@ -305,6 +324,9 @@ def __init__(
             F (float): Sensitivity factor, for a single sensitivity factor calibration
             ms_cal_results (list of MSCalResult): The mass spec calibrations, required
                unless mol, mass and F are given.
+            ms_cal_result_ids (list of int): The id's of the mass spec calibrations,
+               used when loading from a database instead of `ms_cal_results`. The
+               `MSCalResult`s are then only loaded when they are needed.
```
ScottSoren: it is very little memory that this saves, but nice to be consistent about the principle :)

ScottSoren (commented):
[WIP]
I've reviewed all but two modules and played around a bit. It looks and runs fantastic! My comments now are mostly questions for understanding.
I expect to finish the review on the train today or tomorrow.
A question for the meanwhile: assuming we choose in the end to use `sqlalchemy`, is it any more work to start from ixdat-sqllite than it is to start from ixdat-main?

src/ixdat/db.py
```diff
@@ -45,19 +47,32 @@ def save(self, obj):
         """Save a Saveable object with the backend"""
         return self.backend.save(obj)
 
+    @contextmanager
+    def temporary_backend(self, backend=None):
```
ScottSoren, line 51: brilliant.

src/ixdat/db.py
```diff
@@ -111,6 +126,14 @@ class Saveable:
     `extra_column_attrs` to add extra columns via an auxiliary table without changing
     the main table name.
 
+    A class can also say what Python type of value each of its columns holds, with
+    `column_types`, and which columns hold the id of a row of another table, with
+    `column_references`. Relational backends need this to build their tables (see
+    :module:`~ixdat.backends.relational`); backends which just serialize each object
+    on its own, like the directory backend, ignore it. Both are optional: a column
```
ScottSoren: This explanation confuses me a bit. The directory backend also intends to be relational - so maybe it's "SQL-based" or "externally implemented" backends that need explicit typing. Also not clear from here why str/int/float are okay and others aren't. Is it because `assert a == eval(str(a))` works?

src/ixdat/db.py
```diff
@@ -125,9 +148,21 @@ class Saveable:
...
+            ancestry, so each class only declares the tables and columns it adds.
+        extra_linkers (dict): {table_name: (reference_table, attr)} for defining
+            the connections between objects. Also merged over the class ancestry.
+        column_types (dict): {attr: python_type} giving the type of the value stored
+            in a column, for the columns where it matters. Supported types are
+            ``int``, ``float``, ``str``, ``dict``, ``list``, ``tuple``, and
+            ``numpy.ndarray``. A column left out here has no fixed type. Definitions
+            are merged over the inheriting classes, so a class only declares the
+            columns it adds itself.
+        column_references (dict): {attr: table_name} for the columns which hold the
```
ScottSoren: As I understand it, the only reason putting e.g. `"field_id"` in `column_attributes` isn't enough has to do with ensuring that the field is saved when the referencing object gets saved. So I suggest mentioning that here.

src/ixdat/db.py
```diff
@@ -403,16 +477,29 @@ def from_dict(cls, obj_as_dict):
...
+        return DB.get(cls, i, backend=backend)
+
+    @classmethod
+    def load(cls, name, backend=None):
+        """Open the most recently saved object of cls with the given name"""
+        return DB.load(cls, name, backend=backend)
+
+    @deprecate(
+        "0.3.0",
+        "`load_data` takes the backend to load the data from. Pass it as `backend=`.",
```
ScottSoren, line 597: nice catch and correction!

src/ixdat/db.py
```diff
@@ -386,15 +406,69 @@ def save(self, db=None):
...
+    @classmethod
+    def get_column_references(cls):
+        """Return {attr: table_name} for all of cls's columns which hold a foreign id
+
+        Merged over the inheriting classes, exactly like `get_column_types`.
+        """
+        return cls._merged_class_dict("column_references")
+
+    @classmethod
+    def _merged_class_dict(cls, attr):
```
ScottSoren, line 571: very nice tool. I think Kenneth ended up making something very similar for #75.

src/ixdat/db.py
```diff
@@ -201,7 +238,7 @@ def short_identity(self):
         This is (usually) sufficient to tell if two objects refer to the same thing,
         when used together with the class attribute table_name
         """
-        if self.backend is DB.backend:
+        if self.backend == DB.backend:
```
ScottSoren: Interesting correction. Did you run into a case where `self.backend` would equal `DB.backend` without being `DB.backend`?

src/ixdat/backends/sqlite_backend.py
```diff
@@ -0,0 +1,774 @@
...
+        """Initiate the backend, connecting to (and if needed creating) the database
+
+        Args:
+            db_path (Path or str): The path to the SQLite database file. By
+                default it is named after the project and put in ixdat's
+                standard data directory, mirroring the directory backend.
+            directory (Path): The directory for the default database file
+            project_name (str): The project name for the default database file
+        """
+        self._is_memory = str(db_path) == ":memory:"
```
ScottSoren, line 57: This needs explaining...

src/ixdat/backends/sqlite_backend.py
```diff
@@ -0,0 +1,774 @@
...
+        return str(self.db_path if self._is_memory else self._resolved_path)
+
+    def close(self):
+        """Close the connection to the database file"""
+        if not self._closed:
+            self.connection.close()
+            self._closed = True
+
+    def __enter__(self):
+        """Return this backend for use as a context manager."""
```
ScottSoren: is this something general that should go in `BackendBase`?

src/ixdat/backends/sqlite_backend.py
```diff
@@ -0,0 +1,774 @@
...
+            return True
+        if other.__class__ is not self.__class__:
+            return False
+        if self._is_memory or other._is_memory:
+            # Separate ":memory:" connections are separate databases.
+            return False
+        return other._resolved_path == self._resolved_path
+
+    def __hash__(self):
+        """Hash by the database file, so that equal backends hash alike
```
ScottSoren: why would an ixdat session ever have multiple `SQLiteBackend` instances pointing to the same file? 
Would it make sense to make this class a kind of singlet, using `__call__` to ensure that a second initiation with the same file as an existing object just returns that object again? 
I find using `__hash__` potentially more likely to create unintended consequences somewhere down the line.

src/ixdat/backends/sqlite_backend.py
```diff
@@ -0,0 +1,774 @@
...
+        """Atomically save an object and all objects it references.
+
+        Args:
+            obj (Saveable): the object to save
+            force (bool): Whether to force updates if the object is already saved
+            no_updates (bool): Whether to allow updates if the object is already
+                saved. If both force and no_updates are False, the user will be
+                prompted on whether to save.
+        """
+        plan = self._build_save_plan(obj, force=force, no_updates=no_updates)
```
ScottSoren, line 132: very cool code here :)

src/ixdat/backends/sqlite_backend.py
```diff
@@ -0,0 +1,774 @@
...
+                self._ensure_tables(type(planned_obj))
+
+        original_states = []
+        root_result = None
+        try:
+            # Children come before their parents in "plan", so by the time we
+            # insert a row, anything it points to already has its id. Doing all
+            # inserts/updates on one connection makes them one transaction: if
+            # anything below fails, SQLite rolls every row in "plan" back.
+            with self.connection as connection:
```
ScottSoren, line 144: Does this call the context management methods? (I would have expected `with self as connection` to call them...)

src/ixdat/backends/sqlite_backend.py
```diff
@@ -0,0 +1,774 @@
...
+
+    def _build_save_plan(self, root, force, no_updates):
+        """Return unique objects in child-first order with insert/update actions."""
+        plan = []
+        visited = set()  # objects already added to "plan"
+        visiting = set()  # objects currently being visited, i.e. above us in the walk
+        existing_tables = self._existing_tables()
+
+        def visit(obj, updates_forbidden):
+            object_key = id(obj)
```
ScottSoren, line 181: could you mention in a comment that this is [python's id function](https://www.w3schools.com/python/ref_func_id.asp) and not something from sql or ixdat?

src/ixdat/backends/sqlite_backend.py
```diff
@@ -0,0 +1,774 @@
...
+            if object_key in visited:
+                return  # already planned (e.g. two parents sharing one child)
+            if object_key in visiting:
+                # we're already in the middle of visiting this object further up
+                # the call stack, i.e. it refers back to itself somewhere
+                raise DataBaseError(
+                    f"Can't save a cyclic object graph rooted at {root!r}."
+                )
+            visiting.add(object_key)
```
ScottSoren, line 190: why do you need L192 (adding object_key before the child object loop) and L198 (removing object key again)? It's in `plan` that the order really matters anyway, right?

src/ixdat/backends/sqlite_backend.py
```diff
@@ -0,0 +1,774 @@
...
+                )
+            )
+
+        visit(root, updates_forbidden=no_updates)
+        return plan
+
+    def _save_action(self, obj, force, updates_forbidden, existing_tables):
+        """Return ``insert``, ``update``, or ``skip`` for one planned object."""
+        row_exists = False
+        if obj.backend == self and obj.table_name in existing_tables:
```
ScottSoren: ooh, now I don't like this. This line, with the earlier code determining the behavior of `==`, implies that you could have two seperate open connections to the same sqlite file at the same time checking against each other for a green light. I would much prefer if `is` worked here instead.
... does my concern make sense?

src/ixdat/backends/sqlite_backend.py
```diff
@@ -0,0 +1,774 @@
...
+            row_exists = self.connection.execute(
+                f"SELECT 1 FROM {_quote_identifier(obj.table_name)} " 'WHERE "id" = ?',
+                (obj.id,),
+            ).fetchone()
+        if row_exists:
+            if force:
+                return "update"
+            if not updates_forbidden and prompt_for_permission(
```
ScottSoren, line 220: Looks good! I wonder what sqlalchemy would do in this situation?

src/ixdat/backends/sqlite_backend.py
```diff
@@ -0,0 +1,774 @@
...
+                + f" WHERE {_quote_identifier(linker.owner_column)} = ? "
+                'ORDER BY "position"',
+                (i,),
+            ).fetchall()
+            if rows:
+                linked_ids = [row[0] for row in rows]
+                obj_as_dict[linker.id_attr] = (
+                    linked_ids if linker.is_list else linked_ids[0]
+                )
+        obj = cls.from_dict(obj_as_dict)
```
ScottSoren, line 393: is the column `id` in `row` and thus the key `id` in this `obj_as_dict`? If so, does that render redundant `obj.set_id(i)` on L380 below?

src/ixdat/backends/sqlite_backend.py
```diff
@@ -0,0 +1,774 @@
...
+                identities = [identities]
+            for position, identity in enumerate(identities):
+                linked_id = self._dereference(identity)
+                if not isinstance(linked_id, int):
+                    raise DataBaseError(
+                        f"{obj!r} refers to {linker.id_attr}={identity}, which is "
+                        f"not the id of a row saved in {self}. Do the referenced "
+                        "objects live in another backend?"
+                    )
+                connection.execute(
```
ScottSoren, line 312: Can you put in a comment here?  Maybe with an example? I had to scroll all the way up to see that connection is an `sqlite3.connect` object and all the way down to see the string `_insert_sql `makes, and then think hard about how this together builds the right SQL statement.

src/ixdat/backends/sqlite_backend.py
```diff
@@ -0,0 +1,774 @@
...
+        if row is None:
+            raise DataBaseError(
+                f"{self} has no row with id={i} in table '{main_schema.name}'"
+            )
+        obj_as_dict = {
+            column.name: self._decode(column, value)
+            for column, value in zip(columns, row)
+        }
+        for column in main_schema.data_columns:
+            if column.dtype is np.ndarray:
```
ScottSoren: Checking type should work. But I think so far the attribute / column name `data` is reserved exclusively for the potentially heavy numerical data that one might want to load lazily.
In other words, this should probably say `if column.name == 'data'`.

ScottSoren: Makes very little difference now, but I think simplifies things a bit when lazy-loading the data later to use the fact that it always has the same attribute name.

src/ixdat/backends/sqlite_backend.py
```diff
@@ -0,0 +1,774 @@
...
+            (name,),
+        ).fetchone()
+        if row is None:
+            raise DataBaseError(
+                f"{self} has no row named '{name}' in table '{cls.table_name}'"
+            )
+        return self.get(cls, row[0])
+
+    def load_obj_data(self, obj):
+        """Return the numerical data of an object from its ndarray column."""
```
ScottSoren: See comment to L346 above. If lazy loading is applied completely and exclusively to attributes called 'data' (actually, one could be more specific and limit it to 'DataSeries.data'), it saves the typechecking here on L402.

ScottSoren (commented):
[WIP] 
Only missing the last 300-some lines of sqlite_backend. Still looking great!
I expect to finish and give an approval to merge by end of workday tomorrow.

src/ixdat/backends/sqlite_backend.py
```diff
@@ -0,0 +1,774 @@
...
+
+from . import relational
+from .backend_base import BackendBase
+from ..config import config, prompt_for_permission
+from ..exceptions import DataBaseError
+
+
+DATABASE_FILE_SUFFIX = ".sqlite"
+SCHEMA_VERSION = 1
+METADATA_TABLE = "_ixdat_metadata"
```
ScottSoren: this seems a bit misleading. Isn't that just `METADATA_TABLE_NAME`?

src/ixdat/backends/sqlite_backend.py
```diff
@@ -0,0 +1,774 @@
...
+            except (TypeError, ValueError):
+                raise DataBaseError(
+                    f"Invalid ixdat schema version {row[0]!r} in {self.db_path}."
+                )
+            if database_version != SCHEMA_VERSION:
+                relation = "newer" if database_version > SCHEMA_VERSION else "older"
+                raise DataBaseError(
+                    f"The database {self.db_path} uses {relation} ixdat schema "
+                    f"version {database_version}; this ixdat supports version "
+                    f"{SCHEMA_VERSION}. Migrate the database before opening it."
```
ScottSoren: but if relation is "older", then it should be painless to open. Should this just print or prompt a warning in that case?

src/ixdat/backends/sqlite_backend.py
```diff
@@ -0,0 +1,774 @@
...
+    def _ensure_tables(self, cls):
+        """Create, validate, and additively migrate tables needed by ``cls``."""
+        # this only needs doing once per class per backend instance, since a table
+        # doesn't change shape again while this backend stays open:
+        if cls in self._ensured_classes:
+            return
+        existing_tables = self._existing_tables()
+        ensured_definitions = []
+        with self.connection as connection:
+            for schema in relational.table_schemas_of(cls):
```
ScottSoren, line 502: Given that `table_schemas_of` returns the schemas for the the family of tables, how is `_ensure_existing_family_tables` different than this method (other than the input being a class instead of schema)?

src/ixdat/backends/sqlite_backend.py
```diff
@@ -0,0 +1,774 @@
...
+                ):
+                    continue  # table doesn't exist yet, or was already checked
+                self._validate_or_migrate_table(connection, schema)
+                self._ensure_indexes(connection, schema)
+                ensured_definitions.append(schema_definition)
+        self._ensured_schema_definitions.update(ensured_definitions)
+        return existing_tables
+
+    def _validate_or_migrate_table(self, connection, schema):
+        """Add missing nullable columns and reject incompatible table layouts."""
```
ScottSoren: As I think we talked about, I think we should just say that you can't use an ixdat sqlite file with an older version of ixdat than the one that generated it. So then I think this method (and maybe a few of the others) aren't needed in practice?
Moving data from an old schema to a new schema should be supported, but I think that would always go via ixdat objects by means of a `get()` and a `save()`. So all that needs to be supported is getting data from an older version. Perhaps that would also run through this code? But it shouldn't need to... just initiate the objects with the columns available, and the other attributes just get the default values as set in `__init__`?

src/ixdat/backends/sqlite_backend.py
```diff
@@ -0,0 +1,774 @@
...
+        for column in schema.columns:
+            if column.name not in table_info:
+                if column.name == "id":
+                    self._incompatible_schema(
+                        schema.name, 'primary-key column "id" missing'
+                    )
+                # a new column is always safe to add: it's nullable, so existing
+                # rows just get NULL for it - nothing to fill in or guess at.
+                connection.execute(
+                    f"ALTER TABLE {_quote_identifier(schema.name)} ADD COLUMN "
```
ScottSoren: ah, yeah, if I understand this, this is going into the original (old-schema) sqlite file and altering stuff. I don't think this behavior should be accessible. But open for discussion if you've forseen a situation I haven't.

src/ixdat/backends/sqlite_backend.py
```diff
@@ -0,0 +1,774 @@
...
+            project_name = project_name or config.default_project_name
+            self.db_path = directory / (project_name + DATABASE_FILE_SUFFIX)
+        self._resolved_path = None  # what identifies this database, see __eq__
+        if not self._is_memory:
+            self.db_path.parent.mkdir(parents=True, exist_ok=True)
+            # resolved once here: two backends on one file must compare and hash
+            # alike however their paths were written, and resolving hits the disk
+            self._resolved_path = self.db_path.resolve()
+        self.connection = sqlite3.connect(self.db_path)
+        self.connection.execute("PRAGMA foreign_keys = ON")
```
ScottSoren, line 87: Could you put a comment here linking to t[his documentation for PRAGMA](https://system.data.sqlite.org/home/doc/dc206da59f/Doc/Extra/pragma.html), since it's not standard SQL?

src/ixdat/backends/sqlite_backend.py
```diff
@@ -0,0 +1,774 @@
...
+    def _existing_tables(self):
+        """Return the set of names of the tables in the database"""
+        rows = self.connection.execute(
+            "SELECT name FROM sqlite_master WHERE type = 'table'"
+        ).fetchall()
+        return {row[0] for row in rows}
+
+    @staticmethod
+    def schema_report():
+        """Return the CREATE TABLE statements of all imported Saveable classes
```
ScottSoren, line 665: beautiful!

src/ixdat/backends/sqlite_backend.py
```diff
@@ -0,0 +1,774 @@
...
+        if column.dtype in (dict, list, tuple):
+            decoded = json.loads(value)
+            return tuple(decoded) if column.dtype is tuple else decoded
+        return value
+
+    def _dereference(self, value):
+        """Turn a reference to another object into a plain id, for storing as a FK
+
+        An id-holding attribute (e.g. `s_ids`) can hold either a plain int id, or,
+        if the referenced object isn't in this backend yet, a `(backend, id)` pair
```
ScottSoren: You've built on top of something that I thought maybe should have been tor down: the overloaded id type. 
The background is that I hold strongly that we do need support for an ixdat object with references to data in multiple backends. You may want to load one set of data from a database and another from a raw data file and `+` them, or "join" them on a scanned variable like potential, resulting in such a chimera. But it seemed overkill for the backend to be included in all id's when most ixdat sessions will probably be using a single backend... or two if you count the memory backend that data nominally lives in right after read in from raw data files and not yet saved, i.e. 
```
ec = Measurement.read(...)
ec.save()
ms = Measurement.read(...)
ecms = ec + ms   # ecms is now a chimera
ecms.save()  # should not re-save any data series that were saved by ec.save(), ecms no longer a chimera.
```
Then again, that two-backend session (memory plus a sqlite, for example) is a natural enough use that perhaps id's should *always* be the tuple version. 
What do you think?

ScottSoren (commented):
These are my last in-code comments.
I skimmed through the testing modules (they look good) but won't review them in detail.

I still want to:
- re-read and reply to the all the points in your PR description.
- commit and make a little PR with my extended demo'ing

But yeah this is awesome. Let's talk release plan ASAP :)

ScottSoren (approved):
Approving this now to merge after you've looked through my comments, most of which are either very minor or clarification questions for my sake or just compliments. Very happy - as you know from #75, we've been wanting to make an SQL backend for a long time.
I think the benefits would outweigh the extra work of switching to `sqlalchemy`, but I think that we should merge this first! 
For that reason it might not make sense fully perfectionising this SQLite backend - it is good enough to release. The only thing I'd like to get first is the two (of 20) readers that somehow return unsaveable measurement objects (see #212), but we can do that with a small follow-up PR.
Here answers to your questions.

>     Is each piece of information described in one place? A class should list the new information that it saves. Information already described by a parent class should be reused. Please look for places where the same saved value is described more than once.

Yes, I think so! I've spent some time saving stuff and looking through the webapp, and I don't think I see any excess saving of data series (which are the important part). For example: (1) Data series appear to be only saved once even though they are referenced by both combined and component measurements. (2) When spectra are extracted from SEC measurements, they use the same data series for wavelength. 

When a measurement is saved, we've made the decision that it should ensure all its component measurements are saved. This in practice seems to often result in identical metadata in many rows of the measurements table, with just the measureemnt name and series referenced changing. One could revisit that decision to avoid clutter in that table. But I think it's fine.

>     Can ixdat save the kinds of values people normally use? The current list covers whole numbers, decimal numbers, text, dictionaries, lists, tuples, and NumPy arrays. Please consider whether a typical ixdat plugin would need another common kind of value.

I can't think of anything else it should need to save.

>     Do saved objects stay connected correctly? A measurement can point to data series, calculators, component measurements, and spectra. Some of these are ordered lists. Please check that saving and loading keeps every connection and keeps list items in the same order. 

I don't think the order should ever matter. But it doesn't hurt to save and apply it.

>     What should happen when ixdat changes later? If a future ixdat version adds a new optional value, it can add an empty place for that value to an old database automatically. If a change could alter the meaning of existing data, ixdat stops and explains that the database needs a careful manual update. Please check whether that is a sensible starting rule.

We've talked a bit about this, and I did take a bit of issue with the current implementation in a comment. I think "updating" of an sqlite-saved ixdat object would take place by loading it and saving it with ixdat. Then yes, columns would be added, but not as a database migration - it'd be implemented in the ixdat objects themselves. So some of the implementation might be overkill - but that's fine.

Great work! Really looking forward to getting this out in the world :)

development_scripts/demo_sqlite_webapp.py
FrederikLizakJohansen: Nice!

development_scripts/demo_sqlite_backend.py
FrederikLizakJohansen: It does re-save the data series as well. `force=True` currently means “overwrite this measurement and the connected objects it owns". Avoiding only unchanged parts would need a general way to track changes across the full object graph, which I think could grow into a monster implementation very quickly.

development_scripts/demo_sqlite_backend.py
FrederikLizakJohansen: Essentially they are pointing to the same SQLite file, but they are separate connections. One is being used by ixdat to communicate with the db and the other is being used to show (in the demo) that you can make a direct sqlite3 connection to the same db and use ordinary SQL queries against the same file.

development_scripts/demo_sqlite_backend.py
FrederikLizakJohansen: Yes. This is deliberately the same call as at the start. After `DB.backend.close()`, it creates a fresh `SQLiteBackend` and opens a new ixdat connection to the same file.

src/ixdat/backends/memory_backend.py
FrederikLizakJohansen: The difference from this to the `DirBackend` is that `MemoryBackend` does not select rows or reload objects. It already holds the actual python objects in its objects dictionary, so it can return them directly.

We could make name lookup faster by keeping an extra dictionary:  name -> newest id, which would make it not need to scan through the objects, but that also adds some extra bookkeeping whenever an object is saved or renamed.

src/ixdat/backends/relational.py
FrederikLizakJohansen: No it is just an internal name that I chose for the table that an extension table adds values to. However, I agree that it can be confusing, so in 68b9a3309e2e689b79af97cee42cbfdd00043b11 I have changed it to "base_table", so that we avoid "verbifying" or whatever they call it.

src/ixdat/backends/relational.py
FrederikLizakJohansen: You are right that, for measurements, we could first read the main row, use technique to choose `ECMSMeasurement`, and then read only that class’s extension tables. The current family-table scan is not strictly necessary for that case. I used it as one general loading path because subclass selection currently lives separately inside `Measurement.from_dict()`, `DataSeries.from_dict()`, and `Calculator.from_dict()`.

development_scripts/demo_sqlite_webapp.py
FrederikLizakJohansen: I'll take a look at your new demo PR and see if we can make this better there.

src/ixdat/backends/relational.py
FrederikLizakJohansen: Great question! Yes we may need a set later for a plugin perhaps. I did not find a current ixdat column that stores one, and JSON has no native set representation, so adding it requires some extra decisions about ordering and supported element types. The `KNOWN_COLUMN_TYPES` is extendable, so I suggest that we add sets when we have a concrete use case.

src/ixdat/techniques/spectroelectrochemistry.py
FrederikLizakJohansen: Oh! At least we need it for the `ECOpticalMeasurement` loading to be robust.

src/ixdat/techniques/spectroelectrochemistry.py
FrederikLizakJohansen: Yeah, this is probably a design flaw. We should probably lean more up against what sqlalchemy does. As far as I understand, they handles this idea with a foreign key and an explicit relationship. I will take a look at this ASAP.. I might have an idea.

src/ixdat/calculators/ms_calculators.py
FrederikLizakJohansen: Principles are nice right ;)

src/ixdat/db.py
FrederikLizakJohansen: Yeah probably "relational" is too broad, if I understand what you are pointing to correctly. I meant a backend which has to build typed SQL columns, specifically SQLite in this PR. 

As far as I undestand, Python's SQLite driver can store str, int, and float values directly. A dict, list, tuple, or numpy-array needs explicit rules for turning it into a database value and also rebuilding the original python values later again. There is no `eval(str(a))` involved, to my knowledge. I will see to changing the description here to be more accurate.

src/ixdat/db.py
FrederikLizakJohansen: When closing and reopening the same SQLite file, this becomes relevant. Say, an object loaded before the reopen would still hold the old SQLiteBackend instance, but DB.backend would after the reopen be a new instance. `is` would return `False` because they are different objects. `==` would return `True` because both instances points to the same SQLite file. I hope that makes sense :-)

src/ixdat/backends/sqlite_backend.py
FrederikLizakJohansen: You are completely right, I will add some docs here to explain it better.

Basically, as I have understood it,  ":memory:" is a special SQLite address that when used, say like: `sqlite3.connect(":memory:")`, creates just a temporary database inside that one connection with no file on disk. What is special about this address is, that if two instances of a connection is made with the ":memory:" address, they become separate databases that just share the same address (very unlike a regular address). This means that in a case where the user is trying to compare two databases (see `__eq__` below), we need this "_is_memory" to make sure we are comparing correctly.

FrederikLizakJohansen: But taking a second look at this, this implementation actually does not cover the case where we want to compare two simultaneous in-memory databases. I will try make the adjustments in a PR commit.

src/ixdat/backends/sqlite_backend.py
FrederikLizakJohansen: Yes I'll make add this.

src/ixdat/backends/sqlite_backend.py
FrederikLizakJohansen: Yes you are right; if we want to extend to other backends in the future, we should make this be general. I will make the changes in a commit.

src/ixdat/backends/sqlite_backend.py
FrederikLizakJohansen: The main case is closing and reopening the same file. I want that second init to create a fresh connection.

FrederikLizakJohansen: I looked into the `__call__` idea more closely. Defining `__call__` on `SQLiteBac`kend would make an existing backend callable as` backend()`. It would not control calls to `SQLiteBackend("project.sqlite")`; doing that would require a custom metaclass.

But the idea is possible. We could implement the singleton that way, although it would make callers share the same live connection. F.ex: one caller might keep a backend for a longer task while another uses the same file in a short `with` block or something. A singleton would return the same object to both callers, and leaving the `with` block would close the connection for both; that's a big problem. There is also the case of potential multi-threading, where separate threads needs separate SQLite connections.

This is why I think separate backend instances make sense even when they point to the same file. 

`__hash__` does not combine those instances. It is there because `__eq__` needs it. Two backends pointing to the same file compare equal, and python requires equal objects to have the same hash.

src/ixdat/backends/sqlite_backend.py
FrederikLizakJohansen: `with self.connection` calls specifically the context-manager methods of the underlying `sqlite3.Connection`. That is one that commits the transaction if the block succeeds, but rolls it back if an exception happen instead. If we use `with self` we would call the context-manager method on the `SQLiteBackend`, whose `__exit__()` would close the whole backend. We only want to close the sqlite connection after one save and not the whole backend. I hope that makes sense.

src/ixdat/backends/sqlite_backend.py
FrederikLizakJohansen: Yes I will see to doing that; and also explain in a comment why we are using it.

src/ixdat/backends/sqlite_backend.py
FrederikLizakJohansen: `visiting` is used to detect circular relationships while the plan is being built. For example, if A points to B and B points back to A, A is still in visiting when the traversal reaches it again. ixdat can then stop with a clear cycle error instead of recursing forever.

After all of an object’s related objects have been checked, it leaves `visiting` and enters `visited`. `visited` means the object is fully planned and prevents a shared object from being added twice. Say f.ex. if A points to C and B also points to C, then once C has been planned through A, it will be skipped when then afterwards reached through B.

I hope this makes sense.

src/ixdat/backends/sqlite_backend.py
FrederikLizakJohansen: Yeah, your concern is valid, I can see some issue now. My comments on the previous review-comments might not make sense after this.

The solution, I think, is to make a dedicated function called something like "`shares_storage_with()`", that can check the instances directly. An explicit implementation like this would also prevent future contributors from misinterpreting the use of `==`. It would also let us remove the custom `__hash__` that we discussed above, which would be a plus, as well as resolve some of the questions about ":memory:".

I will make an implementation of this for the PR, and get back to this comment.

src/ixdat/backends/sqlite_backend.py
FrederikLizakJohansen: Yeah I don't think it normally prompts.

src/ixdat/backends/sqlite_backend.py
FrederikLizakJohansen: I agree, I'll make some comments.

src/ixdat/backends/sqlite_backend.py
FrederikLizakJohansen: Yes I can see the distinction. I was only concerned that using the name would effect future implementation that would use other strings than "data" for lazy loading; but I get that the whole idea is that "data" is the convention. I will change this in a future commit and get back to your comment.

src/ixdat/backends/sqlite_backend.py
FrederikLizakJohansen: `id` is not included in row. The `WHERE "id" = ?` part uses i to choose the row, while the `SELECT` part controls which columns SQLite returns. `obj_as_dict` therefore contains only the values needed to construct the object, with `data=None `as the lazy-loading signal.

src/ixdat/backends/sqlite_backend.py
FrederikLizakJohansen: Yes, agreed. See my comment above. I will make the implementation.

src/ixdat/backends/sqlite_backend.py
FrederikLizakJohansen: Yes this is the thing that we talked about. I totally get your concerns; and I do think your proposed migration direction is safer.

I suggest that this PR supports version 1 without all the automatic in-place migration mechanisms. We keep validation and table-family discovery, remove the `ALTER TABLE` behavior, and document the future read-old/write-new migration model. I will make the alterations and come back to this comment.

src/ixdat/backends/sqlite_backend.py
FrederikLizakJohansen: You are right, this is too overloaded. I think we should make `short_identity` consistently return (backend, id). I will attempt to incorporate this into a revision of this section and return with the verdict.

src/ixdat/backends/sqlite_backend.py
FrederikLizakJohansen: See https://github.com/ixdat/ixdat/pull/209#discussion_r3813115152

src/ixdat/backends/sqlite_backend.py
FrederikLizakJohansen: Yes good idea; I will make the change in a commit.

src/ixdat/backends/sqlite_backend.py
FrederikLizakJohansen: See https://github.com/ixdat/ixdat/pull/209#discussion_r3813115152

src/ixdat/backends/sqlite_backend.py
FrederikLizakJohansen: See https://github.com/ixdat/ixdat/pull/209#discussion_r3813115152

src/ixdat/db.py
FrederikLizakJohansen: See https://github.com/ixdat/ixdat/pull/209#discussion_r3807445242

src/ixdat/techniques/ms.py
FrederikLizakJohansen: I think this was a mistake from my side, good catch. Watch out for the solution coming in https://github.com/ixdat/ixdat/pull/209#discussion_r3724530654

FrederikLizakJohansen:
@ScottSoren Thanks a lot for the (very) thorough review and the approval! I’ve gone through al the comments now, and will spend the next few days implementing the changes I’ve promised in the individual comments. Once those are ready, I’ll push a new version with a summary of what changed. Some of the updates touch the shared relationship, identity, and schema logic, so I might ask for a quick re-review before we merge. I’ll let you know when it is ready.

FrederikLizakJohansen:
@ScottSoren I have now made the follow-up changes.

The main design change is the new `Relationship` declaration. The idea is to describe each relationship in one place: the stored id, the related Python object, whether the relation points to one object or several, and how it should be saved. This removes the one-item-list workaround for the EC-Optical reference spectrum, and inherited relationship information now follows the same class-order rules as the rest of ixdat's saved information.

I also separated the identity of a backend object from the storage it gives access to. For example, two SQLite connections to the same file are still two different Python objects, while `shares_storage_with()` now tells ixdat that they access the same stored rows. References now use `(backend, id)` consistently (after our suggestion), and `same_short_identity()` handles the case where two separate connections point to the same stored row.

The 'migration of schema' behavior now follows the direction that you suggested. Reading checks the file and leaves it unchanged. A future migration can then read a supported old file and save the resulting ixdat objects into a new file. I also limited delayed array loading to the column named data, moved context management into `BackendBase`, gave each SQLite "`:memory:`" connection its own identity, and added the requested comments around foreign-key setup, Python's id(), and the linker-table insert.

For the full set of post-review changes, you can look at commit [7c46b4a](https://github.com/ixdat/ixdat/pull/209/commits/7c46b4a78db23abcc470eb992d362b761771c208). Also, sorry for the force-pushes. I had some issues with git along the way.

src/ixdat/backends/sqlite_backend.py
FrederikLizakJohansen: see https://github.com/ixdat/ixdat/pull/209#issuecomment-5368055002

src/ixdat/backends/sqlite_backend.py
FrederikLizakJohansen: See https://github.com/ixdat/ixdat/pull/209#issuecomment-5368055002

src/ixdat/backends/sqlite_backend.py
FrederikLizakJohansen: See https://github.com/ixdat/ixdat/pull/209#issuecomment-5368055002

src/ixdat/backends/sqlite_backend.py
FrederikLizakJohansen: See https://github.com/ixdat/ixdat/pull/209#issuecomment-5368055002

src/ixdat/backends/sqlite_backend.py
FrederikLizakJohansen: See https://github.com/ixdat/ixdat/pull/209#issuecomment-5368055002

src/ixdat/backends/sqlite_backend.py
FrederikLizakJohansen: See https://github.com/ixdat/ixdat/pull/209#issuecomment-5368055002

src/ixdat/backends/sqlite_backend.py
FrederikLizakJohansen: See https://github.com/ixdat/ixdat/pull/209#issuecomment-5368055002

src/ixdat/db.py
FrederikLizakJohansen: See https://github.com/ixdat/ixdat/pull/209#issuecomment-5368055002

src/ixdat/backends/sqlite_backend.py
FrederikLizakJohansen: See https://github.com/ixdat/ixdat/pull/209#issuecomment-5368055002

src/ixdat/backends/sqlite_backend.py
FrederikLizakJohansen: See https://github.com/ixdat/ixdat/pull/209#issuecomment-5368055002

src/ixdat/backends/sqlite_backend.py
FrederikLizakJohansen: See https://github.com/ixdat/ixdat/pull/209#issuecomment-5368055002

src/ixdat/backends/sqlite_backend.py
FrederikLizakJohansen: See https://github.com/ixdat/ixdat/pull/209#issuecomment-5368055002

src/ixdat/backends/sqlite_backend.py
FrederikLizakJohansen: See https://github.com/ixdat/ixdat/pull/209#issuecomment-5368055002

src/ixdat/backends/sqlite_backend.py
FrederikLizakJohansen: See https://github.com/ixdat/ixdat/pull/209#issuecomment-5368055002

src/ixdat/backends/sqlite_backend.py
FrederikLizakJohansen: See https://github.com/ixdat/ixdat/pull/209#issuecomment-5368055002

src/ixdat/techniques/spectroelectrochemistry.py
FrederikLizakJohansen: See https://github.com/ixdat/ixdat/pull/209#issuecomment-5368055002

FrederikLizakJohansen: Specifically introduced the Relationship deceleration

FrederikLizakJohansen:
@ScottSoren Thanks for adding the addition to the demo, and the discussion, in #212, you caught something that I had completely not thought about.

I tested the Autolab and Pfeiffer files from your test data. The Autolab reader has text values like `10 µA` in the "Current range" column, and the Pfeiffer reader has formatted date-time strings. Pandas stores those columns as numpy object arrays, which SQLite is refusing to save.

It really comes down to us using `allow_pickle=False`, but we cannot just allow for pickles, because it allow loading a malicious code. So the backend should keeps pickle disabled and only saves arrays with a clear data representation.

My fix: If an object array contains only strings, ixdat converts it to numpys normal unicode array before saving. This should cover the common pandas text-column case from readers and plugins.

Object arrays with mixed python values still raise an error. Cases with dates, dictionaries, `None`, decimals, or mixed values need their own saved representation, because ixdat needs to know what those values mean before choosing how to store them. We need to enforce readers to put their data into clear standardized formats in the future.

Longer term, I think the most defensible solution is to make the `DataSeries.data` contract clear and enforce it where ixdat first receives the data. A `DataSeries` should contain a numpy array with a native numerical dtype. It should not contain arbitrary Python objects. Then `DataSeries.__init__()` can catch object arrays immediately.

For these two readers, that would probably mean converting Autolab values like `1 µA`, `10 µA`, and `100 µA` into numerical amperes, and omitting Pfeiffer’s `Time Absolute (Date_Time)`, because `Time Absolute (UTC)` already contains the same timestamps numerically.

I would still keep a storage-level check in SQLite as the final safety layer. If ixdat eventually needs time-dependent text, we can add an explicit representation for it, for example a `CategoricalSeries` that stores integer codes plus a list mapping those codes to labels. That gives text-like series a defined format without allowing arbitrary Python objects.

I checked the original files after the fix. The problematic readers now saves and reloads

For #212, I think the clean path is to merge #209 first and then retarget #212 to `main`. Then the discussion about demo-data location and paths can continue there.

development_scripts/demo_sqlite_backend.py
ScottSoren: Ah, there's a principle in ixdat that is not enforced in code and is maybe not documented: `DataSeries` are immutable. Anything that changes data in a measurement creates a new `DataSeries`. Thus, an implementation that does not overwrite unchanged objects already in the database would not need to check if they've been modified: it would just overwrite mutable objects like measurements and not touch data series (which is presumably where most of the reading/writing comp time would be). 

This should eventually be coded in a proper way, but not now.

src/ixdat/backends/relational.py
ScottSoren: nounifying here, I think ;) 
Good change.

docs/source/diving_deeper/backend.rst
```diff
@@ -51,54 +82,84 @@ convenient when the lifetime is limited::
...
-*several* rows of another table belongs in ``extra_linkers`` instead.
+SQLite stores containers as JSON and arrays as blobs. As described above, the column
+named ``data`` stays in the file until its property is accessed; other array columns
+load with the object's ordinary values. Pandas represents text columns as object
+arrays. When every item in such an array is a string, SQLite stores it as a native
+NumPy Unicode array so that it can be loaded safely without pickle. A column which is
+not listed has no fixed type and stores its scalar value directly. Saving a container
+or array to an untyped column raises a ``DataBaseError`` naming the column.
+
+The key in ``relationships`` is the Python attribute holding the linked object.
```
ScottSoren, line 145: nice.

src/ixdat/backends/__init__.py
```diff
@@ -21,6 +21,6 @@
 # FIXME: should automate that all initiated backends get added to database_backends?
 database_backends = {
     "none": BackendBase(),  # Just assigns id's but doesn't keep track.
-    "memory": MemoryBackend(),  # Keeps track so child objects can be passed around
+    "memory": MemoryBackend(),  # Keeps related objects available by id
```
ScottSoren, line 24: :)

src/ixdat/backends/backend_base.py
```diff
@@ -27,6 +35,37 @@ def __init__(self):
...
+    def __exit__(self, exc_type, exc_value, traceback):
+        """Release this backend's resources when leaving a context manager."""
+        self.close()
+
+    def shares_storage_with(self, other):
+        """Return whether this backend and ``other`` reach the same saved objects."""
+        return self is other
+
+    def _dereference(self, value):
+        """Return the local integer from a ``(backend, id)`` reference."""
```
ScottSoren: what is a "local integer"? Wouldn't it be clearer to call it the "local id"?

src/ixdat/backends/relational.py
```diff
@@ -207,14 +205,29 @@ def extension_table_schemas(cls):
...
 
 def linker_table_schemas(cls):
-    """Return the linker-table schemas of a Saveable class and its ancestors."""
+    """Return the connection-table schemas of a Saveable class and its ancestors.
+
+    New relationships state ``many`` directly. Older ``extra_linkers`` declarations
+    keep their established rule: an id attribute ending in ``_ids`` contains a list.
+    """
+    many_by_table = {
+        relationship.storage_table: relationship.many
```
ScottSoren: wait, won't all of the values in this dictionary be `True`? What is the point of making it a dictionary rather than a list or set?

src/ixdat/backends/sqlite_backend.py
ScottSoren: Yes, makes sense now. Thanks! 
Could use a comment in the code.

src/ixdat/backends/sqlite_backend.py
```diff
@@ -396,15 +416,19 @@ def load(self, cls, name):
...
-                row = self.connection.execute(
-                    _select_sql(main_schema.name, [column.name]) + ' WHERE "id" = ?',
-                    (obj.id,),
-                ).fetchone()
-                return self._decode(column, row[0]) if row else None
+        data_column = next(
+            (column for column in main_schema.data_columns if column.name == "data"),
+            None,
+        )
+        if data_column is None:
```
ScottSoren: I don't think this should pass silently. The only table with a numerical data column is data_series, and an ixdat object only has any business calling this method if it is an instance of `DataSeries`.
Knowing that context can probably simplify the above SQL querry.

src/ixdat/db.py
```diff
@@ -112,6 +112,60 @@ def get_database_name():
...
+
+    Args:
+        linked_table (str): The table containing the related object.
+        id_attr (str): The attribute which returns the related object's saved id,
+            or its ordered ids when ``many=True``.
+        many (bool): Whether the object attribute contains an ordered list.
+        storage_table (str or None): The table which stores the relationship. A
+            relationship containing many objects requires its own table of
+            connections. A relationship containing one object uses the owner's main
+            table when this is left out.
```
ScottSoren, line 129: What if `many` is `False` but `storage_table` is not `None`? Should that raise an error?

src/ixdat/backends/sqlite_backend.py
ScottSoren: I don't see how that comment answers this specific question?

src/ixdat/calculators/ms_calculators.py
```diff
@@ -58,15 +58,14 @@ def get_bg_for_t(self, t):
 class MSBackgroundSet(Calculator):
     calculator_type = "MS background"
 
-    extra_linkers = {
-        "ms_background_constants": ("ms_constant_backgrounds", "ms_constant_bg_ids")
-        # Additional types of backgrounds would go here
+    relationships = {
```
ScottSoren, line 61: nice.

src/ixdat/techniques/spectroelectrochemistry.py
```diff
@@ -58,8 +58,11 @@ class ECOpticalMeasurement(SpectroECMeasurement):
 
     default_plotter = ECOpticalPlotter
 
-    extra_linkers = {"ec_optical_measurements": ("spectrums", "ref_id")}
-    child_attrs = SpectroECMeasurement.child_attrs + ["reference_spectrum_list"]
+    relationships = {
```
ScottSoren, line 61: nice.

src/ixdat/data_series.py
```diff
@@ -136,15 +139,27 @@ def __init__(self, name, unit_name, data, a_ids=None, axes_series=None):
...
+        return identity
 
     def get_axis_series(self, axis_number):
         """Return the DataSeries of the `axis_number`'th axis of the data"""
         if not self._axes_series[axis_number]:
-            self._axes_series[axis_number] = DataSeries.get(i=self._a_ids[axis_number])
+            identity = self._a_ids[axis_number]
+            if isinstance(identity, int):
+                backend = (
+                    self.db.backend if self.backend_type == "none" else self.backend
```
ScottSoren, line 157: This line looks weird. If this data series doesn't have a backend, how does it have a db?
Is `self.db` the globally active database, controlled by `ixdat.db.change_database()`? 
This would read more logically if so:
```
backend = self.backend or DB.backend
```
...
I see I created this problem by giving `Saveable` a class attribute called `db`. I think now, that was a mistake. It looks wrong for a saveable object to have both its own backend and a reference to the global backend. 
I can see another problem with the above line, which is that `self.backend` is `BackendBase` by default, not `None`. I wonder if that is really necessary? 
Let's revisit this if it is still relevant after sqlalchemy integration.

src/ixdat/db.py
ScottSoren, line 229: Would deleting this, and importing `DB` from `ixdat.db` where needed instead, cause any problems? See https://github.com/ixdat/ixdat/pull/209/changes/2345600204708eb92f08799490c8f1a96030e511..648c74f6749f9a3ae2902c9a0722d91e452007c8#r3966413833

src/ixdat/measurement_base.py
```diff
@@ -50,13 +56,26 @@ class Measurement(Saveable):
         "sample_name": str,
         "tstamp": float,
     }
-    extra_linkers = {
-        "component_measurements": ("measurement", "m_ids"),
-        "measurement_calculators": ("calculator", "c_ids"),
-        "measurement_series": ("data_series", "s_ids"),
+    relationships = {
```
ScottSoren, line 59: nice. This becomes much more explicit.

ScottSoren (approved):
Nice work! This makes relationships much more explicit.

I tested it by merging the updates into #212 , which still works well. 
Reading the code led to a few more comments and questions, including unresolving a couple of the older ones. 

A general discomfort that it adds more code than it eliminates. In that sense, it's not a simplification of our home-made object relationship mapping. Hopefully sqlalchemy integration will be a net reduction in ixdat code.

Could you 
- make the few unresolved comments that aren't easily resolved (by just an answer or a comment in the code) into Issues?
- and then merge it?

src/ixdat/backends/backend_base.py
FrederikLizakJohansen: Yes, local id is clearer. I changed the wording in b5f6f98

src/ixdat/backends/relational.py
FrederikLizakJohansen: Yes, all the values were `True`. This code only needs to know whether the table is present, so I changed it to a set called `many_tables` in b5f6f98.

src/ixdat/backends/sqlite_backend.py
FrederikLizakJohansen: Agreed. `load_obj_data()` now explicitly requires a `DataSeries`, reads its data column, and raises an error if the row is missing. This is in b5f6f98.

src/ixdat/db.py
FrederikLizakJohansen: Intentionally not actually. It describes one related object whose id is stored in an extension table. Say for example `ECOpticalMeasurement` stores one `ref_id` in `ec_optical_measurements`, and `SpectroMeasurement` stores one `spectrum_id` in `spectro_measurements`. A relationship with `many=True ` needs a separate connection table for its list, which is why that is the combination that requires `storage_table`.

src/ixdat/data_series.py
FrederikLizakJohansen: #215

src/ixdat/db.py
FrederikLizakJohansen: idk, it may affect plugins or other external code that accesses self.db, so that needs a compatibility check first. The core uses can likely import DB explicitly.

Issue is active: #215

FrederikLizakJohansen:
Thanks! I addressed the four small code and comment points in `b5f6f98`, answered the two remaining questions in their threads, and opened #214 and #215  for the larger follow-up work. I’ll merge now.

src/ixdat/db.py
ScottSoren: Not 100% with you. I think one of us is confusing or combining the concepts of `storage_table` and `connection_table`, but I'm not sure who ;)

src/ixdat/data_series.py
ScottSoren: thanks!

development_scripts/demo_sqlite_backend.py
ScottSoren: https://github.com/ixdat/ixdat/issues/214 . Thanks for making the issue!

PR #197, opened 2025-08-29
Developing the SEC module in ixdat

Steven687548:
Added a reader for Oceanview time-series output files.
Changes were made to make EC + optical work.
Added a class called StaircaseSEC for processed SEC data (e.g., averaged spectra at potential steps), and its plotter.

PR #212, opened 2026-08-17
Demo tools for sqlite

ScottSoren:
This is just to share a couple ideas about demo scripts and what I'm using to play around with the sqlite backend, for discussion. 
It modifies the webapp demo to a function called `view_tables`. A few minor problems came up:
- the autolab and pfeiffer readers somehow put in some object in a numpy array that can't be put into sqlite without pickling. I haven't looked into it yet.
- (not related to sql) Some of the plotters use a deprecated (now-broken) syntax for matplotlib color maps: the spectroelectrochemistry waterfall plot (which I think will be fixed in #197 ), and also `plot_cycles`.

In addition, this PR is a minor step on the way to a better demo data organization. I think the `DEMO_DATA_DIR` variable should be in a config rather than in a python script like it is now, but couldn't figure out the best way. What do you think is the best way to have large data files for testing? (We once played around with git submodules, but I think it might be easier to have people download an archive from Erda or another non-git cloud file storage service.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 
 # Big files from tests and docs prep
+development_scripts/**/*.sqlite
 development_scripts/**/*.png
 development_scripts/**/*.svg
 development_scripts/**/*.csv
--- a/development_scripts/reader_demonstrators/demo_autolab_reader.py
+++ b/development_scripts/reader_demonstrators/demo_autolab_reader.py
@@ -1,12 +1,9 @@
 """For use in development of the autolab reader. Requires access to sample data."""
 
-from pathlib import Path
-
+from tools_for_demos import DEMO_DATA_DIR
 from ixdat import Measurement
 
-path_to_file = Path.home() / (
-    "Dropbox/ixdat_resources/test_data/autolab/autolab_test_file.txt"
-)
+path_to_file = DEMO_DATA_DIR / "autolab/autolab_test_file.txt"
 
 meas = Measurement.read(path_to_file, reader="autolab")
 
--- a/development_scripts/reader_demonstrators/demo_avantage_reader.py
+++ b/development_scripts/reader_demonstrators/demo_avantage_reader.py
@@ -1,12 +1,9 @@
 """Sandbox script to aid in development and demo of the Avantage reader"""
 
-from pathlib import Path
-
+from tools_for_demos import DEMO_DATA_DIR
 from ixdat import Spectrum
 
-path_to_file = (
-    Path.home() / "Dropbox/ixdat_resources/test_data/avantage" / "XPS Survey.avg"
-)
+path_to_file = DEMO_DATA_DIR / "avantage/XPS Survey.avg"
 
 spectrum = Spectrum.read(path_to_file, reader="avantage")
 
--- a/development_scripts/reader_demonstrators/demo_biologic_mpr_reader.py
+++ b/development_scripts/reader_demonstrators/demo_biologic_mpr_reader.py
@@ -1,12 +1,15 @@
-from pathlib import Path
+from tools_for_demos import DEMO_DATA_DIR
 from ixdat import Measurement
 
-for data_dir in [
-    Path.home() / "Dropbox/ixdat_resources/test_data/biologic/17J04_Pt_isotope_exchange",
-    Path.home() / "Dropbox/ixdat_resources/test_data/biologic/22I27_London",
-    Path.home() / "Dropbox/ixdat_resources/test_data/biologic/22K14_Tempo",
+combined_measurement_list = []
+
+for folder in [
+    "17J04_Pt_isotope_exchange",
+    "22I27_London",
+    "22K14_Tempo",
 ]:
     combined_meas = None
+    data_dir = DEMO_DATA_DIR / "biologic" / folder
 
     for file in data_dir.iterdir():
         if not file.suffix == ".mpr":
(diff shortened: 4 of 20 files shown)


FrederikLizakJohansen:
A little aside, but relevant to our discussions on standardizing the demo scripts across ixdat: I suggest that we introduce (keep) the `main()` function in all maintained demo scripts.

The demo has two jobs as I see it. First, someone should be able to run it directly and see the examples (say reviewers of our code). Second, another script should be able to reuse the example data. The SQLite demo tool here is one such script.

I suggest a general system like:

```python
def load_example():
    # Create or load the objects used by the example.
    ...

def main():
    # Run the visible demonstration.
    ...

if __name__ == "__main__":
   main()
```

How we actually load the data in a generalizable way; I am still considering.

FrederikLizakJohansen:
Extending on my previous comment, I have thought of a plan to standardize the demo scripts:

1) As has already been discussed, I think we should  put the files used by the reader demos in one ZIP archive somewhere anyone can download it, maybe definitely ERDA. Keep the folders familiar (`autolab/...`, `pfeiffer/...`, and so on). We'll need to check that we're allowed to share the files and swap out any that we aren't.

2) Then we keep one tiny `.ini` file in the repo, `demo_data.ini`, that says which archive to get:
```ini
  [archive]
  version = v1
  url = <direct download URL>
  sha256 = <checksum of the ZIP file>
```
  
3) We then use [Pooch](https://www.fatiando.org/pooch/latest/about.html) to handle the download, checksum, cache, and unzip step. We can give people one command: `python download_demo_data.py`. The useful bit of that script is basically:
```python
  pooch.retrieve(
      url=config["archive"]["url"],
      known_hash="sha256:" + config["archive"]["sha256"],
      fname="ixdat-demo-data-{}.zip".format(config["archive"]["version"]),
      path=cache_dir,
      processor=pooch.Unzip(extract_dir=config["archive"]["version"]),
  )
``` 
We'd document `python -m pip install pooch` for people running the demos. The script would read `demo_data.ini` with python's `configparser`.

4) By default, the data can live in Pooch's usual cache folder. Anyone who wants it somewhere else can set `IXDAT_DEMO_CACHE_DIR`. A helper in `tools_for_demos.py` would return that folder plus the archive version, so every demo looks in the same place. This keeps personal paths out of the scripts.

5) We then (by the suggested convention I mentioned above) would give each demo a `load_example()` that reads the data and a `main()` that makes plots or other visible output. Then the SQLite script can import the loader without accidentally opening plots. Something like:
```python
  def load_example(data_dir=None):
      root = Path(data_dir) if data_dir is not None else get_demo_data_dir()
      return Measurement.read(root / "autolab/autolab_test_file.txt", reader="autolab")

  def main():
      load_example().plot()

  if __name__ == "__main__":
      main()
```

ScottSoren:
Nice find with pooch! It seems we are exactly who it was made for :) 


Okay - with the `load_example` function, then I see the point of the `if __name__ == "__main__"` structure. With the following caveat (let me know if you agree):

`load_example` should have a docstring saying what it returns, should return exactly what one might want to play around with in an ipython console, and should cache it.

For example, demo_autolab_reader.py becomes
```
from pathlib import Path
from ixdat import Measurement
from ixdat.demos import get_demo_data_dir


loaded_data = {}

def load_example(data_dir=None):
    """return an autolab measurement by reading a single test file"""

    root = Path(data_dir) if data_dir is not None else get_demo_data_dir()
    meas = Measurement.read(root / "autolab/autolab_test_file.txt", reader="autolab")
    loaded_data["example_1"] = meas 
    return meas

def load_example_2(data_dir=None, use_cache=True):
    """return an autolab measurement by reading and combining a CP and CVA file"""
    if use_cache and "example_2" in _data:
        return data["example_2"]

    root = Path(data_dir) if data_dir is not None else get_demo_data_dir()
    meas = Measurement.read_set(root / "autolab/autolab_set", reader="autolab", suffix=".txt")
    loaded_data["example_2"] = meas
    return meas

def main():
    meas_1 = load_example()
    meas_1.plot()
    meas_2 = load_example_2()
    meas_2.plot()
    meas_full = meas_1 + meas_2
    meas_full.plot()

if __name__ == "__main__": 
    main()
```
The cache'ing could, I think, easily be implemented in a decorator.

Another script could get the data it needs in a controlled way:
```
from pathlib import Path
from ixdat.db import change_db()
from ixdat.demos.demo_autolab_reader import load_example
from ixdat.demos.demo_tools import view_tables


sqlite_file = Path.home() / "desktop/test.sqlite"
change_db("sqlite", db_path=sqlite_file)

m1 = load_example()
m1.save()
view_tables(sqlite_file)
```

A spyder or pycharm Ipython console could look like:
```
In [1]: %runfile  demo_autolab_reader.py ...
[output, plots, ...]
In [2]: m = loaded_data["example_2"]   # it was just there waiting for me :) 
In [3]: m.calibrate(R_Ohm=100)   # or whatever the developer, reviewer, curious user wants to check out.
In [4]: m.plot_vs_potential()
>>>
```

As you can tell I've made a couple assumptions in the above demos that should be questions. 
- How do you import from a demo script? Relative or absolute? Above I've assumed that the demos live in src/ixdat, so that they can be imported as part of the ixdat package.
- Where does the function get_demo_data_dir() live? I've put it in a src/ixdat/demos/__init__.py (maybe via import from demo_tools.py)

